PR #7, opened 2023-05-24
Don't reinstall flutter if it's already available

jupblb:
I modified direnv instructions so that `flutter` is not installed if it's already available.


jupblb:
Merging as it doesn't contain any changes outside of devenv.

PR #13, opened 2023-06-02
Cleanup

jupblb:
This PR is a product of browsing the code based on my UI journey and deleting all the useless widgets. I've also removed a few comments that did not bring much value.

It does make browsing the app/codebase a bit easier for me so I think that's a benefit. I'm against putting more time into this as we are going to replace the views anyway (most likely).


jupblb:
Done

shadowbrush (approved):

PR #14, opened 2023-06-05
Mock server updates for mentor request iteration

andersan:
Work in progress. Will add any generators or states needed for the mock server updates in the mentor request iteration.

Not sure if any additional changes are needed, will check with @shadowbrush on this in a bit.

### DIFF
--- a/mm-mock-server/src/mocks/mutations.ts
+++ b/mm-mock-server/src/mocks/mutations.ts
@@ -1,4 +1,5 @@
 import { MockServerState } from "./util/state";
+import * as generators from "./util/generators";
 
 export function mockMutations(serverState: MockServerState) {
     return {
@@ -14,6 +15,34 @@ export function mockMutations(serverState: MockServerState) {
         signMeOut: () => {
             serverState.loggedIn = false;
             return "ok";
+        },
+        createChannel: () => {
+            const channel = generators.generateChannel([serverState.loggedInUser, serverState.otherUsers[1]])
+            serverState.channels.push(channel);
+            return channel;
+        },
+        createChannelInvitation: () => {
+            const invitation = generators.generateChannelInvitation([serverState.loggedInUser, serverState.otherUsers[1]])
+            serverState.channelInvitations.push(invitation);
+            return invitation;
+        },
+        createChannelMessage: () => {
+            const message = generators.generateChannelMessage(serverState.loggedInUser, true)
+            serverState.channelMessages.push(message);
+            return message;
+        },
+        updateChannelInvitation: (status: any) => {
+            const invitation = serverState.channelInvitations[0]
+            if (status) {
+                if (status === "accepted"){
+                    invitation.status = "accepted";
+                    invitation.channel = generators.generateChannel([serverState.loggedInUser, serverState.otherUsers[3]])
+                    serverState.channels.push(invitation.channel);
+                }
+                else if (status === "declined")
+                    invitation.status = "declined";
+            }
+            return invitation.id;
         }
     }
 }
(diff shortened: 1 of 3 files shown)


shadowbrush:
The idea here is that specifying an existing channel is optional. Specifying the sender and recipient should suffice. The channel would be created when the recipient accepts the invitation using `updateChannelInvitation`, setting the `status` to `ChannelInvitationStatus.accepted`. But reviewing the code, it needs work. I didn't put much time into this, as I didn't know whether the app would need invitations. Now that I know it does, I'll finish the implementation.

shadowbrush:
I update the BgChannels code to not require a `channelId` when calling `createChannelInvitation`. There is a test that verifies that `updateChannelInvitation` creates a new 1:1 channel if the invitation didn't specify an existing channel: https://github.com/baragaun/channels-service/blob/d3064c25b0279c05f8256c42a889b937ae254f76/src/services/bgChannels/test/models/channelInvitation/updateChannelInvitation.spec.ts#L60

andersan:
@shadowbrush the `Channel` still doesn't seem to be created. [Added a PR in that repo](https://github.com/baragaun/channels-service/pull/11) which may fix the issue, but I wasn't sure how to run tests within the repo so I'm not sure that it's bug-free.

andersan:
Ah I just reread your message above and I see that the `Channel` isn't created at all until the invitation is accepted. Feel free to close that PR if that's the case - I had thought the Channel would need to exist along with the invitation.

andersan:
FYI pausing on this work for the moment. I was getting caught up on how to load all the data into the mock server that we need to implement the first set of flows. We may need to think about adding new types or modifying existing types in the `mmdata` backend.

Some definitions for the purpose of this PR:
1. A "Request" is a request to connect. The sender can be a mentor or mentee.
2. A "Match" is a request to connect that has been accepted.

In `mmdata`, "Request" is a `ChannelInvitation` which has a status of `created`. All pending `Request`s sent to a User will be available via the `MyInbox` GraphQL query.

A "Request" is accepted and becomes a "Match" via the `updateChannelInvitation` GraphQL mutation which sets `ChannelInvitation.status` to `accepted`. However, we can't just look at whether the `ChannelInvitation.status` field to see if there is a new match. **I assume** we only want to display **new** Matches to the user in the Inbox screen. It's not clear if we want to display a "Match" in a user's inbox if they are the one who accepted the invitation, or what the flow after accepting will look like.

In our backend, I think this is currently the way we'd check for a "Match":
1. Get a list of `Channel`s that this user is a part of
2. Check if any of those `Channel`s don't have any messages exchanged

Not sure how we'd do this in the mock GraphQL server, so that's why I'm stopping here. I may strip down this PR since much of the added complexity was attempting to get the "Match" added.

The Inbox screen isn't part of this iteration, but was included in the first set of flows, so maybe this isn't so urgent.

shadowbrush:
> Ah I just reread your message above and I see that the Channel isn't created at all until the invitation is accepted. Feel free to close that PR if that's the case - I had thought the Channel would need to exist along with the invitation.

Exactly. The invitation can exist without a channel. I'll close your PR, and left a comment there, too.

andersan:
Still not sure about what constitutes a Match, but I don't think this is important in the short term.

@m-rosario could you take a look at this and let me know what you think? I want to make sure the changes to the state don't introduce too much complexity.

m-rosario (approved):
Thanks for adding these!

andersan:
Thanks Miguel, added these to `main`. I'll add more changes to the mock server for the next iteration. There are some questions I still need answered from the Product/UX side before those get finalized.

PR #15, opened 2023-06-07
First iteration of dashboard (mock)

m-rosario:
Replaces the old Home screen with a prototype of the new dashboard/home page with very basic styling and almost no functionality. Adds several widgets and styling decisions that can be reused in other parts of the application in order to maintain consistency. Specially important are the constants introduced, which encapsulate themes, styles, dimensions, and other design choices.


lib/utilities/reminder_banner.dart
```diff
@@ -0,0 +1,41 @@
...
+
+class ReminderBanner {
+  static void showReminderBanner(
+      BuildContext context, Icon? icon, String text, Function() onOpenPressed) {
+    ScaffoldMessenger.of(context).showMaterialBanner(
+      MaterialBanner(
+        padding: const EdgeInsets.all(16),
+        content: Text(text),
+        leading: icon,
+        backgroundColor: Colors.white70,
```
LucasMagnum, line 13: Totally optional: what do you think about abstracting the colors in a Theme object? Then we can easily changed that later on without having to change in the widget.

m-rosario: Yes absolutely! Using an App level Theme will make a lot of these declarations simpler. At this stage I am just quickly putting stuff together to see what works and what doesn't. Once we see the patterns in colors and styles they should be abstracted into constants / themes. I can take care of doing this next

LucasMagnum (approved):

PR #16, opened 2023-06-08
Create navigation bar and replace some navigator logic with go_router logic

ashlilili:
1. Created widget navigation_bar
2. Created file test_main, as the entry point of the testing app under development 
3. Disabled widgetbook enabled go router in package specs
4. Replaced signin screen navigator with router


lib/test_main.dart
```diff
@@ -0,0 +1,166 @@
...
+    return userProvider.queryUser(
+      onLoading: () {
+        return const LoadingScreen();
+      },
+      onError: (error) {
+        return const SignInScreen();
+      },
+      onData: (data) {
+        WidgetsBinding.instance.addPostFrameCallback((_) {
+          GoRouter.of(context).go('/home');
```
jupblb, line 100: We can use `context.go('/home')`

lib/widgets/atoms/navigation_bar.dart
```diff
@@ -0,0 +1,99 @@
+import 'package:flutter/material.dart';
+import 'package:go_router/go_router.dart';
+
+class CustomizedBottomNavigationBar extends StatefulWidget {
```
jupblb, line 4: Since this widget defines a `Scaffold` (along with e.g. top app bar) I'd consider renaming it to somethign else. Like `CustomizedScaffold` or `AppWrapper`.

lib/widgets/atoms/navigation_bar.dart
```diff
@@ -0,0 +1,99 @@
...
+  const CustomizedBottomNavigationBar({Key? key, required this.child})
```
jupblb, line 5: We should write constructor in a way that uses `super.key`. This can be replaced with: `const CustomizedBottomNavigationBar({super.key, required this.child});`

lib/widgets/atoms/navigation_bar.dart
```diff
@@ -0,0 +1,99 @@
...
+      return 0;
+    }
+    return 0;
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      appBar: AppBar(
+          backgroundColor: const Color.fromARGB(255, 255, 255, 255),
```
jupblb, line 58: You may use `Colors.white` here.

jupblb (changes requested):
Regarding https://github.com/micromentor-team/mm-flutter-app/pull/16/commits/3075695671a629553708710605d0c6056a0f3db5 this should be definitely done. I wonder if it shouldn't be in another PR but I'll leave this decision with you. Either way @jonahtwalker should be notified about it.

As this is the code that we want to work with in the future I'd rename current `main.dart` to `main.old.dart` (or sth like that) and rename `test_main.dart` from this PR to `main.dart`.

Absolutely awesome job with both writing entire `router` definition and building `CustomizedBottomNavigationBar`. I do believe atm we can drop the micromentor logo on top (this does not match the design views).

We will also want to avoid having a `Scaffold` inside another `Scaffold` but this is out of scope for this PR. I do believe having it defined only at `CustomizedBottomNavigationBar` is the way to go.

I left a few other notes in the code diff, mostly nitpicks.

lib/widgets/atoms/navigation_bar.dart
ashlilili: Done!

lib/widgets/atoms/navigation_bar.dart
ashlilili: Done!

lib/widgets/atoms/navigation_bar.dart
ashlilili: Makes sense, changed to AppWrapper

lib/test_main.dart
ashlilili: Done! Thanks a lot for the review!

jupblb:
I still believe we don't need the MicroMentor image at the top of the page but otherwise LGTM

ashlilili:
> I still believe we don't need the MicroMentor image at the top of the page but otherwise LGTM

Oh sorry I forgot to reply here, I included it bc its in the UX design but we can change it later

PR #19, opened 2023-06-13
Added an analysis on design patterns

shadowbrush:
This is a single markdown document discussing the choice of a design pattern.


LucasMagnum:
I do personally like the feature centred architecture. Even though I understand the benefits of granular components (atoms), it feels like an over specialisation for me (YAGNI). 

I vote for using the Feature-Based architecture, I'd consider evaluating the atomic design later on in the project (perhaps after the fellowship).

LucasMagnum (approved):

jupblb:
I agree with Lucas. At this moment we should maximize our initial velocity of developing the product. This can be refactored later.

andersan:
Feature based pattern looks good to me.

What are the implications for a design kit? I had imagined that `atoms` would contain things like different kinds of buttons, a date dropdown, filterable multiselect input, and other low-level components that will be used across the entire app. For now, adding these to the relevant feature is good. Most of them will probably only be used in one feature (e.g. onboarding and editing a profile for inputs). Later, maybe we will have a `shared` folder at the feature level (same as `feature_1`, `feature_2`, etc.).

I'm also not great at Flutter, so maybe I'm missing that such low-level widgets are not necessary. Can we just stick to using Material widgets, or will we need many customized low-level widgets? 

feel free to ignore if this seems to be a distraction from the goal of this PR

shadowbrush:
I added your comments and the poll. This can be merged after it was reviewed.

LucasMagnum (approved):

PR #20, opened 2023-06-13
Use context.push  to enable back button navigation

LucasMagnum:
When using context.go in the BottomNavigationBar it will close the app if you try to press the back button because there is nothing to "pop". Using context.push will push the routes into the navigation stack and when pressing the back button it pop the most recent route instead of closing the app.


LucasMagnum:
> Makes sense. LGTM.
> 
> One thing to check: if a view is pushed on stack does it mean that the previous view is kept in the rendering buffer? Let me give an example. Suppose two screens, A and B, depend on the same `Provider provider`. I move from screen A to screen B using `push`. If `provider`'s state changes, does it mean that both screen A and B will be rerendered? My intuition says "no, only B will be rerendered" but I remember having some flaky experience with that. If my memory is correct then we should ask the Flutter team about that.

Valid point. I am not sure how smart flutter is when it comes to the rendering state. As we will need review the state management soon, I will merge this PR and we can double check it later.

PR #22, opened 2023-06-13
Add explore view

jupblb:
This is a draft of "Explore" view. Things missing at this moment:
* Filtering (search phrase works)
* Quick match
* Entrepreneurs view

We may consider merging this as is and add tickets for missing functionality.

Additional questions/notes to the UX:
1. The quick match button is at awkward position. It is not affected by search or filtering.
2. What should the entrepreneurs view look like? Can we have a wireframe for that?
3. What should a user list item look like? What should be visible before and after expanding (tapping it)? For now I'm reusing `Users` widget in this PR.
4. The search mechanism forces refresh after every character. This is very resource consuming, especially w. r. t. network usage. We should have a button confirmation instead.
5. "Find a mentor" label does not make sense being in the middle of the view. The entire view is about finding a mentor.

I'd personally recommend removing "Find a mentor" rectangle and having two buttons next to a search box: "Search" and "Quick match" ("I feel lucky" 😄).

Additional questions to the backend team:
1. Does the backend support filters for querying user list?
2. How does handling of entrepreneurs work? What is a difference between a mentor and entrepreneur? Does it support the same operations (filtering, search, etc.)?

@LucasMagnum @shadowbrush

### DIFF
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -3,5 +3,14 @@
     "navExploreText": "Explore",
     "navJourneyText": "Journey",
     "navInboxText": "Inbox",
-    "navProfileText": "Profile"
+    "navProfileText": "Profile",
+    "mentors": "Mentors",
+    "entrepeneurs": "Entrepeneurs",
+    "india": "India",
+    "marketing": "Marketing",
+    "online": "Online",
+    "filters": "Show filters",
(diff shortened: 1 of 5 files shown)


jupblb:
I'll merge it as is. Let's wait for wireframes before making more changes.

PR #25, opened 2023-06-13
Quick fix on loading screen

ashlilili:
signin crashes the app when there's errors (eg. the input user doesn't exist in the database), the problem is scaffoldManager.showSnackBar in signin screen requires that a scaffold always exists


LucasMagnum (approved):
Thanks for catching this.

PR #29, opened 2023-06-16
Reminder banner UI

ashlilili:
Added the UI for reminder banner.
Currently supports 3 scenarios based on PRD: 
1. when the profile completion percentage is < 50%
completeprofile

2. when the profile hasn't been updated in 6 months
needupdate

3. no banner if neither (1) or (2)
nobanner

====
Small change: removed app bar micromentor icon based on new discussion


lib/widgets/atoms/reminder_banner.dart
```diff
@@ -0,0 +1,125 @@
...
+            const SizedBox(
+                child : Image(
+                image: AssetImage("assets/images/StockImage1.png"),
+                height: Dimensions.reminderBannerImageHeight,
+            )),
+            InkWell(
+            onTap:  () => {
+              debugPrint("hi")
+            },
+            child: Row(children: [
```
m-rosario: The text and icon in this Row are aligned to the start of the row, whereas the AssetImage is always centered. This means that as the screen size increases the image and the text become misaligned. I think the easiest fix is to add MainAxisAlignment.center to the Row, but you can experiment with other approaches

lib/widgets/atoms/reminder_banner.dart
```diff
@@ -0,0 +1,125 @@
...
+
+  Widget _buildImageAndButtonColumn(context){
+    return Expanded(
+      flex: 1,
+      child: Column(
+          mainAxisSize: MainAxisSize.min,
+          children: <Widget> [
+            const SizedBox(
+                child : Image(
+                image: AssetImage("assets/images/StockImage1.png"),
```
m-rosario: Is this image just a placeholder or are we intended to use this in the end product? If so can you link to where you got it?

lib/widgets/atoms/reminder_banner.dart
```diff
@@ -0,0 +1,125 @@
...
+  Widget build(BuildContext context) {
+    return Padding(
+      padding: const EdgeInsets.symmetric(
+        horizontal: Insets.widgetMediumInset,
+        vertical: Insets.widgetSmallestInset,
+      ),
+      child: Card(
+        child: Padding(
+          padding: const EdgeInsets.symmetric(
+        horizontal: Insets.widgetMediumInset,
```
m-rosario: If all sides use the same Inset you can use EdgeInsets.all(Insets.widgetMediumInset)

lib/widgets/screens/dashboard/dashboard.dart
```diff
@@ -50,6 +51,10 @@ class DashboardScreen extends StatelessWidget {
                   profileMessage: _getGreeting(l10n, user),
                   profileCompletionPercentage: 1.00,
                 ),
+                MaybeReminderBanner(
+                  profileCompletionPercentage: 30,
```
m-rosario: Note that the profileCompletionPercentage for the profile progress indicator is a double from 0.0 to 1.00, whereas the reminder banner is using an int from 0 to 100. Either approach is fine but we shouldn't mix both. You can go ahead and change the logic in the progress indicator if you think an int makes more sense. I just chose that because that is what the default widget uses.

lib/widgets/atoms/reminder_banner.dart
```diff
@@ -0,0 +1,125 @@
...
+                height: Dimensions.reminderBannerImageHeight,
+            )),
+            InkWell(
+            onTap:  () => {
+              debugPrint("hi")
+            },
+            child: Row(children: [
+                const Icon(Icons.arrow_forward),
+                Text( ctaText, 
+                style: TextStyles.cardTitle(context)),
```
m-rosario: I like putting a comma at the end of each parameter which also allows the formatter to do it's job. Check out the official docs for their take on this https://docs.flutter.dev/tools/formatting#using-trailing-commas

m-rosario (changes requested):
This is looking great I love it! 👍

Just left a few small comments in the code diff, no big deal :)

Thanks for taking care of this!

lib/widgets/atoms/reminder_banner.dart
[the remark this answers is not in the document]
ashlilili: Ah great idea, a separate file makes it a lot cleaner, thanks!

lib/widgets/atoms/reminder_banner.dart
```diff
@@ -0,0 +1,125 @@
...
+    required this.profileCompletionPercentage,
+    required this.lastUpdateTime
+  }) : super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
+    print(DateTime.now().difference(lastUpdateTime).inDays);
+    if(profileCompletionPercentage < 50){
+      return ReminderBanner(
+        titleText: "Profile $profileCompletionPercentage% complete",
```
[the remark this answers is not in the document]
ashlilili: Added under app_en! Thanks for handling localization

lib/widgets/atoms/reminder_banner.dart
ashlilili: Good catch, fixed

lib/widgets/atoms/reminder_banner.dart
ashlilili: done!

lib/widgets/screens/dashboard/dashboard.dart
ashlilili: Changed the profile icon to using integer

lib/widgets/atoms/reminder_banner.dart
ashlilili: good point, thanks for the pointer

m-rosario (approved):
Thanks!!

PR #30, opened 2023-06-20
Firebase integration with Crashlytics. Error handling, logging, & reporting

m-rosario:
Sets up the foundation for integrating Firebase into the application. The first Firebase service being used is Crashlytics, which will allow for error reporting in release builds.

Additionally, this commits creates the basic infrastructure for error handling by offering several fallback handlers, logging, and reporting. A custom ErrorWidget, Dialogs class, and a retry function are also included as utilities that can be leveraged by developers.


lib/utilities/errors/crash_handler.dart
```diff
@@ -0,0 +1,127 @@
...
+  bool _isCollectionEnabled = false;
+
+  CrashHandler(bool isReleaseMode, bool isCollectionEnabled) {
+    _isReleaseMode = isReleaseMode;
+    _isCollectionEnabled = isCollectionEnabled;
+    _initialize();
+  }
+
+  void _initialize() async {
+    await Firebase.initializeApp();
```
ashlilili, line 24: I pulled from this branch but failed at this step with web, fixed by changing it to `Firebase.initializeApp(options: DefaultFirebaseOptions.currentPlatform);`.

ashlilili (approved):
Super cool stuff! Thanks for adding the firebase integration & error handling logic. I'm excited to use crashHandler in the future, especially the retrying & snackbar logic like the demo we saw!

lib/utilities/errors/crash_handler.dart
m-rosario: Ah very nice catch! I just add a commit to fix this. Thanks!

ashlilili (approved):
LGTM. Thanks for updating github actions!

PR #31, opened 2023-06-21
Invitations section UI with 'See all' and 'See less' functionality

m-rosario:
Removes the previous Match section and replaces with the Invitations section, with UI modified to match latest UX design. Some essential building blocks were modified as part of this PR, notably the SectionTile, which can now display a button that toggles between 'Show all' and 'Show less'.

Screenshot_20230621_160936

Screenshot_20230621_160946


lib/widgets/molecules/invitation_section.dart
```diff
@@ -0,0 +1,123 @@
...
+      ),
+      _Invitation(
+        userName: 'Catalina Echevarria',
+        userJobTitle: 'COO - ACME Corporation',
+        invitationType: InvitationTileType.accepted,
+        avatarUrl:
+            'https://media.istockphoto.com/id/1437816897/photo/business-woman-manager-or-human-resources-portrait-for-career-success-company-we-are-hiring.jpg?b=1&s=170667a&w=0&k=20&c=YQ_j83pg9fB-HWOd1Qur3_kBmG_ot_hZty8pvoFkr6A=',
+      ),
+    ];
+    return invitations;
```
LucasMagnum, line 85: What happens if the list of invitations is too long and we click in see all? All the objects will be loaded in memory and we will have a huge scroll?

m-rosario: Yes that's what will happen with the current implementation. In this PR I was mostly focused on on the UI and implementing a stateful SectionTile with a callback for 'See all' and 'See less'.

I need to confirm with UX if we want to open the list in a different page, as an overlay, etc. At that point I would also implement pagination.

ashlilili (approved):
(jumping in as lucas will be off next week)
Looks good to me, thanks Miguel!

PR #32, opened 2023-06-22
Channel invitation mock server updates

andersan:
Add findChannelInvitationsForUser.
Minor updates to state, myInbox, findUsers.

Goal is to get the API calls in for this iteration. 

For the invitations UI, we can use `findChannelInvitationsForUser` and `findUsers`.
The query for the invitations should return these fields:

```
senderId
recipientId
channelId
id
status
createdAt
```

The query for users should return these fields:
```
fullName
avatarUrl
```

The source for the text shown alongside the invitation, e.g. "CEO - Enron" is not yet defined. It might be the job title and the name of their company, which would require other updates to the mock server, but it's not clear yet.

---

For the overall Home page, we can use `getAuthenticatedUser` to get the user's first name and avatar URL.

### DIFF
--- a/mm-mock-server/src/mocks/queries.ts
+++ b/mm-mock-server/src/mocks/queries.ts
@@ -18,11 +18,16 @@ export function mockQueries(serverState: MockServerState) {
                 generators.generateChannel([serverState.loggedInUser, serverState.otherUsers[0]]),
             ]
         },
-        findUsers: () => {
+        findChannelInvitationsForUser: () => {
             return [
-                serverState.loggedInUser,
-                serverState.otherUsers[0],
+                generators.generateChannelInvitation(serverState.loggedInUser, serverState.otherUsers[0], false, false),
+                generators.generateChannelInvitation(serverState.loggedInUser, serverState.otherUsers[0], false, true),
+                generators.generateChannelInvitation(serverState.otherUsers[0], serverState.loggedInUser, false, false),
+                generators.generateChannelInvitation(serverState.otherUsers[0], serverState.loggedInUser, false, true),
             ]
+        },
+        findUsers: () => {
+            return serverState.otherUsers.concat([serverState.loggedInUser]);
         },
         getAuthenticatedUser: () => {
             if (!serverState.loggedIn) {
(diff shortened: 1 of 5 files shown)


mm-mock-server/src/mocks/queries.ts
```diff
@@ -18,12 +18,17 @@ export function mockQueries(serverState: MockServerState) {
                 generators.generateChannel([serverState.loggedInUser, serverState.otherUsers[0]]),
             ]
         },
-        findUsers: () => {
+        findChannelInvitationsForUser: () => {
```
shadowbrush, line 21: I'm trying to decide whether it makes sense to have this field. 

The backend has a similar field, `findPendingChannelInvitationsForUser`, which returns the all channel invitation that the user has sent out or received and that haven't yet been accepted or declined. 

The backend does have `findChannelInvitationsForUser`, but it returns all invitations, including those that have already been accepted or declined. Not sure why the app would need that.

Even `findPendingChannelInvitationsForUser` may not be needed, as the pending channel invitations are listed in `UserInbox`. It should be possible to pull the user's inbox and retrieve from there everything the app needs to show a listing of new items, including invitations and new messages.

mm-mock-server/src/mocks/util/generators.ts
```diff
@@ -113,15 +116,15 @@ export function generateChannelInvitation(channelParticipants: any[], declined?:
         __typename: "ChannelInvitation",
         id: faker.string.alphanumeric({length: 24}),
         channelName: faker.lorem.words(2),
-        channel: generateChannel(channelParticipants),
+        channel: channel,
```
shadowbrush, line 119: You could shorten this to `channel,`, just like `recipient,` in line 125 or `sender,` in line 127

andersan:
Going to address Holger's feedback, removing `findChannelInvitationsForUser` and relying on the inbox. After some discussion, I understand that the `UserInbox` exposed at the `myInbox` GraphQL query is the right place for this. The `UserInbox` is meant to manage all app notifications for a user, not just messages or invitations, and it's intended to be the standard place for this type of information to be fetched.

mm-mock-server/src/mocks/queries.ts
```diff
@@ -18,12 +18,17 @@ export function mockQueries(serverState: MockServerState) {
...
+                generators.generateChannelInvitation(serverState.otherUsers[0], serverState.loggedInUser, false, false),
+                generators.generateChannelInvitation(serverState.otherUsers[0], serverState.loggedInUser, false, true),
             ]
         },
+        findUsers: () => {
+            return serverState.otherUsers.concat([serverState.loggedInUser]);
+        },
         getAuthenticatedUser: () => {
             if (!serverState.loggedIn) {
                 return null;
```
andersan, line 34: This doesn't match the signature in the backend that we've implemented. If a logged out user runs `getAuthenticatedUser`, there is an error rather than a null response.

@shadowbrush this is [a question Miguel had](https://trello.com/c/KeSnySPu/49-servererror-when-launching-app-without-being-logged-in) for us yesterday: do we expose any way on the backend to check whether a user is logged in without catching an error? At the moment, I didn't see any, and it feels like we can probably avoid running into this error by checking the logged in state on the front end, and only fetching a user if the front end "thinks" that the user is logged in.

andersan:
Ready for another review. I'll add some more PRs with other mock backend changes in addition to this one.

andersan:
Rebased again with main. The commits on this branch should be squashed before merging to main.

mm-mock-server/src/mocks/queries.ts
shadowbrush: I responded to the Trello card, Chris. I'd say we can leave it as it is right now, but the backend needs to throw a better exception.

shadowbrush (approved):

andersan:
This seems to have already been implemented in #46, I'll close this since it just adds clutter/confusion.

PR #34, opened 2023-06-22
Backend for reminder banner

ashlilili:
1. Add profile completion percentage and last update in mock server user schema, add an api getUserProfileInfo  for getting these two fields
2. The app queries mock backend for user info and renders the banner accordingly
3. (small fix) FirebaseCrashlytics doesn't support web app, so I added an extra condition 

next step would be to update onError to use crashHandler.

### DIFF
--- a/lib/data/models/user/user_api.dart
+++ b/lib/data/models/user/user_api.dart
@@ -59,3 +59,12 @@ const String kGetAllUsers = """
     }
   }
 """;
+
+const String kGetUserProfileInfo = """
+  query Q {
+    getUserProfileInfo(){
+      profileCompletionPercentage
+      lastUpdateTime
+    }
+  }
+""";
--- a/mm-mock-server/src/mocks/util/generators.ts
+++ b/mm-mock-server/src/mocks/util/generators.ts
@@ -100,9 +100,7 @@ export function generateUserProfileCompletionPercentage() {
 
 export function generateUserLastUpdateTime() {
     // Returns a random date time from one year before to now
-    var startDate = (new Date(new Date().setFullYear(new Date().getFullYear() - 1))).valueOf(); 
-    var endDate = (new Date()).valueOf();
-    return new Date(startDate + Math.random() * (endDate - startDate)); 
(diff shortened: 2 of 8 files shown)


mm-mock-server/src/mocks/util/generators.ts
```diff
@@ -92,6 +92,20 @@ export function generateChannelInboxItemMessage(channelId: string, sender: any)
...
+    // Returns a random integer between 0 and 100
+    return  Math.floor(Math.random() * 100); 
+}
+
+
+export function generateUserLastUpdateTime() {
+    // Returns a random date time from one year before to now
+    var startDate = (new Date(new Date().setFullYear(new Date().getFullYear() - 1))).valueOf(); 
+    var endDate = (new Date()).valueOf();
+    return new Date(startDate + Math.random() * (endDate - startDate));
```
m-rosario, line 105: Nit: I believe the faker library can help with generating this random data with less code.

m-rosario (approved):
Looks great, thanks!

mm-mock-server/src/mocks/util/generators.ts
ashlilili: Great tip! Changed to faker library.

m-rosario (approved):

andersan (commented):
Still had this tab opened and realized i never sent these review comments from yesterday. Already implemented a fix in #38 and implemented a mock for profile completion percentage on the User resolver in the mmdata backend, too.

lib/data/models/user/user_api.dart
```diff
@@ -59,3 +59,12 @@ const String kGetAllUsers = """
     }
   }
 """;
+
+const String kGetUserProfileInfo = """
+  query Q {
+    getUserProfileInfo(){
```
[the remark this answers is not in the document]
ashlilili: I see , in this PR I added the two fields under User. I agree that we should reuse updatedAt if it's doing the same thing as lastUpdateTime, and add the field in the User resolver! Thanks for catching this

PR #36, opened 2023-06-26
Refactor GraphQL user queries and data models

m-rosario:
The purpose of this PR is to introduce some useful OOP design patterns, strong typing, and structure into the existing GraphQL query operations for _User_. The remaining models and operations will be later refactored in the same manner if this PR is accepted.

The big motivation behind this PR is the gotcha of using GraphQL, which is that queries can return any number of fields, and so a single model class (e.g. UserModel) cannot accurately represent the response received from the server. Some queries will populate some fields of the UserModel, some will populate others, but very rarely you will have an instance with all of the fields available. Meaning that all fields of the model must be nullable, and that caller must be extremely careful to access only the fields that contain data.

Having each query return a very specific result object with only the fields that they need removes this problem. It also makes the providers easier to use since the caller only has access to what they need, instead of having to manage a JSON object and extracting fields directly.

### Key changes

- **BaseProvider** abstract class that is extended by other provider classes. It contains a runQuery function that takes a BaseOperation as input, and returns a BaseResult as output.
- **BaseOperation** abstract class that represent a single GraphQL query. It contains the logic to convert a GQL json response into the corresponding BaseResult object.
- **BaseResult** abstract class that represents the response of a GQL query. Contains only the exact fields that the query can return, with proper typing and nullable fields.
- **Automatic query retries** while remaining on the 'Loading' state. Once the maximum number of attempts is reached the state changes to 'Error'.
- **Refetch** function callback is exposed to the caller to facilitate manually retrying (or refreshing) from the onData callback, without triggering the operation all over again.
- **FetchMore** function callback is exposed to callers to facilitate pagination of data. This might be revisited later to add more abstraction.

### Next steps (if this is approved)

- Implement a similar structure for Mutations
- Implement for all data models and operations
- Explore ways to make better use of the ChangeNotifiers, perhaps by breaking them up into smaller pieces and leveraging the notify function

### DIFF
--- a/lib/data/models/user/queries/get_authenticated_user.dart
+++ b/lib/data/models/user/queries/get_authenticated_user.dart
@@ -1,0 +1,60 @@
+import 'package:mm_flutter_app/data/models/base/base_operation.dart';
+
+class GetAuthenticatedUser implements BaseOperation {
+  @override
+  String get operation => 'getAuthenticatedUser';
+
+  @override
+  String get gql => """
+  query Q {
+    getAuthenticatedUser {
+      id
+      email
+      fullName
+      avatarUrl
+      userHandle
+    }
+  }
+""";
+
+  @override
+  Map<String, dynamic>? get variables => null;
+
+  @override
+  GetAuthenticatedUserResult resultFromResponseData(Map<String, dynamic> data) {
+    return GetAuthenticatedUserResult.fromJson(data[operation]);
+  }
+}
+
+class GetAuthenticatedUserResult extends BaseResult {
+  String id;
+  String? email;
+  String? fullName;
+  String? avatarUrl;
+  String userHandle;
+
+  GetAuthenticatedUserResult({
+    required this.id,
+    this.email,
+    this.fullName,
+    this.avatarUrl,
+    required this.userHandle,
+  });
+
+  GetAuthenticatedUserResult.fromJson(Map<String, dynamic> json)
+      : id = json['id'],
+        email = json['email'],
+        fullName = json['fullName'],
+        avatarUrl = json['avatarUrl'],
+        userHandle = json['userHandle'];
+
+  Map<String, dynamic> toJson() {
+    return {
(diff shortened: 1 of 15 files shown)


lib/data/models/user/queries/get_authenticated_user.dart
```diff
@@ -0,0 +1,60 @@
+import 'package:mm_flutter_app/data/models/base/base_operation.dart';
+
+class GetAuthenticatedUser implements BaseOperation {
```
andersan, line 3: We might not want these to map 1:1 from dart queries to backend queries.

Some places, we may want to get a lot more data or different fields. For example, if I'm editing my own profile, most of the fields directly on the query will need to be returned. Several additional fields (countryOfResidence, ethnicity, firstName, lastName, birthYear, cityOfResidence, spokenLanguages, websites, etc.) will need to be returned here which generally won't be needed in other places where the `GetAuthenticatedUser` query is made.

Would we plan on adding a new `class GetAuthenticatedUserFullProfile` or some similar logic for the "edit profile" functionality? Or does this negate the benefits of standardizing these queries across the app, and we should just add all of the necessary fields to one `GetAuthenticatedUser` class?

andersan (approved):
The patterns used here look good to me. Including default behavior for refetch/retry, API error handling, and pagination in the base provider and then implementing custom logic in individual providers seems like a good approach.

My inline comment about not mapping these queries 1:1 from the flutter app to the backend is something I'd like us to have in mind as we add more UI, where we may use the same queries or mutations but for different purposes.

lib/data/models/user/queries/get_authenticated_user.dart
m-rosario: Yes I was thinking about this as well and I agree that we could create additional classes for different versions of the same query. However, I imagine we would only do this for when there is a huge difference in the amount of data returned between the versions, as was your example.

PR #37, opened 2023-06-27
Home UI Improvement

ashlilili:
Updated the home page to match wireframe design.
1. updated upcoming sessions component according to wireframe
2. changed top profile bar to match wireframe design
3. moved search bar to explore page
4. removed extra sections on homepage
5. added extra button under recommended mentor section
6. updated bottom navigation bar icons

screenshot: 
Screenshot 2023-06-26 at 9 38 03 PM


lib/constants/dimensions.dart
```diff
@@ -35,7 +35,7 @@ class Dimensions {
   static const double appBarHeight = 32.0;
   static const double minimumInteractionWidgetLength = 48.0;
   static const double recommendedMentorTileWidth = 112.0;
-  static const Size upcomingSessionTile = Size(184.0, 96.0);
+  static const Size upcomingSessionTile = Size(80.0, 150.0);
```
m-rosario, line 38: Not super important at this point, but when defining sizes always try to choose a number that is a multiple of 8 preferably, or by 4 in cases you need more precision (e.g. smaller elements). This is to give the illusion of all elements being neatly organized in a grid.

Check out this for more details: https://m2.material.io/design/layout/spacing-methods.html#baseline-grid

lib/widgets/molecules/upcoming_session_tile.dart
```diff
@@ -1,46 +1,45 @@
...
+  //         text: l10n.homeUpcomingSessionJoin,
+  //       ),
+  //     ));
+  //   }
+  //   buttons.add(RectangleButton(
+  //     onPressed: () => {},
+  //     text: l10n.homeUpcomingSessionReschedule,
+  //   ));
+  //   return buttons;
+  // }
```
m-rosario, line 37: Feel free to delete stuff that is no longer needed. There is nothing groundbreaking here and we can always go back to previous versions if we need to.

m-rosario (dismissed):
Very nice, left a couple of in-line nits. Thanks!

lib/widgets/molecules/upcoming_session_tile.dart
ashlilili: Done!

lib/constants/dimensions.dart
ashlilili: Ah great tips! Changed the constants. Thanks!

m-rosario (approved):
👍👍

PR #38, opened 2023-06-27
Mock server profile, getAuthenticatedUser, and graphql schema updates

andersan:
This PR introduces 4 minor changes:
1. Move profile completion logic to the User in the mock backend
2. ~~Don't return `null` for `getAuthenticatedUser` when the user is logged out. Do we want the mock backend to throw an error in this case in order to match the mmdata backend functionality? See [this trello task](https://trello.com/c/KeSnySPu/49-servererror-when-launching-app-without-being-logged-in) for more context.~~ no longer needed
3. Bulk updates to graphql schema, copied in directly from the Apollo Server ui.
4. ~~In the `myInbox` query, return channel messages with a channelId stored in `channels` in the `state.ts` file. I don't know if this is really necessary.~~ this was implemented in #46

### DIFF
--- a/mm-mock-server/src/mocks/queries.ts
+++ b/mm-mock-server/src/mocks/queries.ts
@@ -27,13 +27,6 @@ export function mockQueries(serverState: MockServerState) {
             }
             return serverState.loggedInUser;
         },
-        getUserProfileInfo: () => {
-            return {
-                __typename: "User",
-                profileCompletionPercentage: serverState.loggedInUser.profileCompletionPercentage,
-                lastUpdateTime: serverState.loggedInUser.lastUpdateTime,
-            };
-        },
         myInbox: () => {
             var mockChannelId = faker.string.alphanumeric({length: 24});
             var mockInvitations = [
--- a/mm-mock-server/src/mocks/queries.ts
+++ b/mm-mock-server/src/mocks/queries.ts
@@ -18,14 +18,16 @@ export function mockQueries(serverState: MockServerState) {
                 generators.generateChannel([serverState.loggedInUser, serverState.otherUsers[0]]),
             ]
         },
-        findUsers: () => {
+        findChannelInvitationsForUser: () => {
             return [
-                serverState.loggedInUser,
-                serverState.otherUsers[0],
-                serverState.otherUsers[1],
-                serverState.otherUsers[2],
-                serverState.otherUsers[3],
+                generators.generateChannelInvitation(serverState.loggedInUser, serverState.otherUsers[0], false, false),
+                generators.generateChannelInvitation(serverState.loggedInUser, serverState.otherUsers[0], false, true),
+                generators.generateChannelInvitation(serverState.otherUsers[0], serverState.loggedInUser, false, false),
+                generators.generateChannelInvitation(serverState.otherUsers[0], serverState.loggedInUser, false, true),
             ]
+        },
+        findUsers: () => {
+            return serverState.otherUsers.concat([serverState.loggedInUser]);
         },
         getAuthenticatedUser: () => {
             if (!serverState.loggedIn) {
@@ -63,6 +65,6 @@ export function mockQueries(serverState: MockServerState) {
                     ],
                 }
             }
-        }
+        },
     }
 }
--- a/mm-mock-server/src/mocks/queries.ts
+++ b/mm-mock-server/src/mocks/queries.ts
@@ -18,14 +18,6 @@ export function mockQueries(serverState: MockServerState) {
                 generators.generateChannel([serverState.loggedInUser, serverState.otherUsers[0]]),
             ]
         },
-        findChannelInvitationsForUser: () => {
-            return [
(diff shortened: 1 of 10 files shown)


mm-mock-server/schema.graphql
ashlilili, line 1: Seems like the profileCompletionPercentage field under User added in https://github.com/micromentor-team/mm-flutter-app/pull/34 was removed - this would cause the app to fail because the api kGetUserProfileInfo , which is being used for reminder banner, assumes this field profileCompletionPercentage under User.

mm-mock-server/src/mocks/queries.ts
```diff
@@ -22,20 +22,10 @@ export function mockQueries(serverState: MockServerState) {
             return serverState.otherUsers.concat([serverState.loggedInUser]);
         },
         getAuthenticatedUser: () => {
-            if (!serverState.loggedIn) {
-                return null;
-            }
             return serverState.loggedInUser;
         },
-        getUserProfileInfo: () => {
```
ashlilili, line 30: If we remove getUserProfileInfo, we should change kGetUserProfileInfo under lib/data/models/user/user_api.dart to querying getAuthenticatedUser?

ashlilili (commented):
Thanks for catching this again and cleaning up the backend changes!

mm-mock-server/src/mocks/queries.ts
```diff
@@ -22,20 +22,10 @@ export function mockQueries(serverState: MockServerState) {
             return serverState.otherUsers.concat([serverState.loggedInUser]);
         },
         getAuthenticatedUser: () => {
-            if (!serverState.loggedIn) {
-                return null;
-            }
```
m-rosario, line 27: I believe the real backend returns null when the user is not logged in, aside from throwing an exception. From what Holger mentioned I believe this means that now we can always expect a User and we will instead verify some other field (or more specific exception) to determine if the user is authorized.

m-rosario (approved):

mm-mock-server/src/mocks/queries.ts
andersan: You're right, didn't realize that the backend was returning `data: null` on errors.

```json
{"errors":[{"message":"Cannot return null for non-nullable field Query.getAuthenticatedUser.","locations":[{"line":2,"column":3}],"path":["getAuthenticatedUser"],"extensions":{"code":"INTERNAL_SERVER_ERROR","stacktrace":["Error: Cannot return null for non-nullable field Query.getAuthenticatedUser.","    at completeValue (/Users/chris/Documents/mm3/mmdata/node_modules/graphql/execution/execute.js:594:13)","    at /Users/chris/Documents/mm3/mmdata/node_modules/graphql/execution/execute.js:486:9","    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)","    at async Promise.all (index 0)","    at async execute (file:///Users/chris/Documents/mm3/mmdata/node_modules/@apollo/server/dist/esm/requestPipeline.js:220:37)","    at async processGraphQLRequest (file:///Users/chris/Documents/mm3/mmdata/node_modules/@apollo/server/dist/esm/requestPipeline.js:155:32)","    at async internalExecuteOperation (file:///Users/chris/Documents/mm3/mmdata/node_modules/@apollo/server/dist/esm/ApolloServer.js:585:16)","    at async runHttpQuery (file:///Users/chris/Documents/mm3/mmdata/node_modules/@apollo/server/dist/esm/runHttpQuery.js:129:29)","    at async runPotentiallyBatchedHttpQuery (file:///Users/chris/Documents/mm3/mmdata/node_modules/@apollo/server/dist/esm/httpBatching.js:34:16)","    at async ApolloServer.executeHTTPGraphQLRequest (file:///Users/chris/Documents/mm3/mmdata/node_modules/@apollo/server/dist/esm/ApolloServer.js:496:20)"]}}],"data":null}
```

mm-mock-server/src/mocks/queries.ts
andersan: Yes, made this change in the latest updates

mm-mock-server/schema.graphql
andersan: Good catch, Ashley. I pulled these graphql updates from a remote apollo server rather than my latest local. This is fixed in the latest commit on this PR.

andersan:
Will fix dart formatting locally in the morning and then this should be ready for another review. I added `updatedAt` to the `getAuthenticatedUser` response and added that to the User provider Flutter reminder widget. Flutter is building for me locally and I can run the app in chrome with these changes.

ashlilili (dismissed):
Thanks Chris!

andersan:
@m-rosario could you review this again when you have a chance? I rebased with main after the changes over the past few days.

Thanks!

m-rosario (approved):

andersan:
The tests were appearing as pending when I tried to merge this. There was some bug where I had to refresh this PR page to see the results which I was already seeing [in Github actions](https://github.com/micromentor-team/mm-flutter-app/actions/runs/5487635611).

PR #40, opened 2023-06-28
Use GraphQL Codegen to create model classes for User queries/mutations

m-rosario:
This introduces the use of the [GraphQL Codegen](https://pub.dev/documentation/graphql_codegen/latest/) package to generate model classes for queries/mutations, as defined in the schema.graphql and operations.graphql files.

This build upon the previous iteration which introduced the use of unique model classes for each GraphQL operation. By completely automating the creation of model classes it will now take a few lines of code to run a GraphQL operation, and receive its result in a properly typed object.

### Key Changes

- Should eliminate the need to manually create model classes for GraphQL operations.
- We can easily have multiple versions of the same query (with different fields returned), since the codegen will take care of creating the model classes for each version with their corresponding fields.
- Will eventually make everything under the data/models directory obsolete, except for the provider classes. At that point these might be moved to a different directory structure.
- All operations now return a OperationResult<T> object, which contains the model class that represents the output of the operation.
- BaseOperation and BaseResult (from previous iteration) are now fully replaced with generated classes. The main logic for executing GraphQL operations remains mostly unchanged.


andersan:
Looks good on an initial skim - in the morning I'll take a more thorough look at this. I can see it negates the need for some of the changes I added in #38 to add `profileCompletionPercentage` and `updatedAt` to the get_authenticated_user query. I will rebase that PR on this one once this is approved and merged into main.

jupblb (changes requested):
This is an important and highly beneficial change. We must keep in mind, however, that the `graphql_codegen` dependency is maintained by a single developer, is only somewhat popular (>100 stars) and actively seeks [sponsorship](https://github.com/heftapp/graphql_codegen/tree/main/packages/graphql_codegen#help-us-survive-become-a-sponsor) for ["improving user productivity and accessibility through AI"](https://github.com/sponsors/heftapp).

lib/data/models/base/base_provider.dart
```diff
@@ -9,10 +9,15 @@ import 'package:retry/retry.dart';
 import '../../../widgets/atoms/server_error.dart';
 
 abstract class BaseProvider extends ChangeNotifier {
-  Widget runQuery<T extends BaseResult>({
-    required BaseOperation operation,
+  final GraphQLClient client;
+
+  BaseProvider({required this.client});
+
+  Widget runQuery({
```
andersan, line 16: Really separate from the scope of this PR so including as a single comment: should we use [Flutter Hooks](https://medium.com/graphql-flutter/graphql-flutter-hooks-23ab5e77514e) rather than Widgets for API queries?

android/app/build.gradle
```diff
@@ -48,7 +48,7 @@ android {
         applicationId "org.micromentor.mm_flutter_app"
         // You can update the following values to match your application needs.
         // For more information, see: https://docs.flutter.dev/deployment/android#reviewing-the-gradle-build-configuration.
-        minSdkVersion flutter.minSdkVersion
+        minSdkVersion 19
```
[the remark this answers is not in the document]
m-rosario: Some of the new dependencies require a min sdk of 19. The flutter min is 16 by default, which is [extremely old](https://apilevels.com/). In fact Im bumping it up to version 21 since less than 1% of devices are using SDKs that low anyway. Also I believe we need to support newer versions to publish to Google Play.

lib/schema/schema.graphql
```diff
@@ -0,0 +1,2302 @@
+type GroupMembership implements IGroupMembership {
```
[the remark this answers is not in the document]
m-rosario: I got the updated schema from Chris, who got it from their Apollo UI. This is more of a backend thing so I will defer to them for suggestions.

lib/data/models/user/user_provider.dart
```diff
@@ -1,24 +1,22 @@
...
 
-class UserProvider extends BaseProvider {
-  GraphQLClient client;
+import '../../../__generated/schema/operations.graphql.dart';
+import '../../../__generated/schema/schema.graphql.dart';
+import '../base/operation_result.dart';
 
-  User? _user;
+class UserProvider extends BaseProvider {
+  Query$GetAuthenticatedUser$getAuthenticatedUser? _user;
```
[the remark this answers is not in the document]
m-rosario: From what I understand Fragments are defined in the schema, which is in the backend (not in our scope). I agree that these are very long, but I see why they went with something like that. We could just use a dynamic variable whenever the type can be easily deducted from the usage (widgets and other classes that call this code).

lib/schema/schema.graphql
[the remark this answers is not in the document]
andersan: In #38 I have the latest backend schema from `mmdata`, not sure if it's exactly the same as here. I also can't guarantee that it's entirely correct. For userHandle, you're right that it should not be nullable, but it is nullable in our backend at the moment. I'll open a PR to mark that non-nullable and can update in this repo later. 

This schema was generated by running the backend locally and downloading this file from the local [Apollo Server playground](http://127.0.0.1:3000/mmdata/api/graphql), in the Schema/SDL view.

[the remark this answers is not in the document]
andersan: FYI it may be possible that we have a "user" that hasn't actually gone through signup (e.g. if we create a "user" for someone once they land on the signup page and accept creation of cookies). I don't know if this is the intended functionality of this collection, but we need to make some decision there prior to modifying this schema to require fields like userHandle, email, etc.

lib/widgets/screens/messages/messages_list.dart
```diff
@@ -1,14 +1,14 @@
...
 import 'package:provider/provider.dart';
 
 import '../../../data/models/channels/channel.dart';
 import '../../../data/models/messages/messages_provider.dart';
-import '../../../data/models/user/user.dart';
 import '../channel_messages/channel_messages.dart';
 
 class MessagesList extends StatelessWidget {
-  final User user;
+  final Query$GetAuthenticatedUser$getAuthenticatedUser user;
```
andersan, line 11: Would the changes we need to make to classes here have implications about code reusability throughout the project? I imagine there will be places we want to display a list of users in some way, but the list of users could come from at least 2 different GraphQL Queries.

In other words, if we build a widget to show a list of Mentors which we get via the `findUsers` endpoint, can we also use that widget to display a list of Mentors which is retrieved via the  `findUserSearchById.topFoundUsers` endpoint? Both of them return a list of users. 

If there's not a clean way to do that, one side effect of this PR would be increased code duplication. If we'd need to make a `BaseMessagesList`, `FindUsersMessagesList`, and `findUserSearchByIdTopFoundUsersMessagesList` widget, that sounds like a lot of overhead.

lib/schema/operations.graphql
```diff
@@ -0,0 +1,56 @@
...
+    getAuthenticatedUser {
+        id
+        email
+        fullName
+        avatarUrl
+        userHandle
+    }
+}
+
+query GetUserProfileInfo {
```
andersan, line 11: I'll remove this and incorporate `profileCompletionPercentage` into the `getAuthenticatedUser` query after we merge this, in PR #38

lib/data/models/user/user_provider.dart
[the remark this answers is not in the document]
andersan: Agreed with this, it would be nice for the Flutter representation of a user variable to look more like `final User user` as it does currently, rather than `final Query$GetAuthenticatedUser$getAuthenticatedUser user`. I don't know how to reconcile this with the changes in this PR, though, so I don't have any suggestion for that.

lib/widgets/screens/messages/messages_list.dart
andersan: @m-rosario what do you think about this? Is there an easy way to deal with this, e.g. by creating a User class with constructors which can take `Query$FindUsers$user` or `Query$findUserSearchById.topFoundUsers.user` as inputs?

m-rosario: Hey @andersan those are great points, here are my two cents.

If we make a generic model to store the value from different queries we would run into the same problem we had before where each query would use different fields of the models (forcing us to make non-shared fields nullable), meaning that the code that uses that model has to make a lot of safety checks.

A query should return a query result object, I think we can all agree on this. The classes that use these results can convert it into anything else they want and pass it around, however if they are consuming the data right there then why bother putting it into another object?

For your example, assuming that both queries return the exact same fields, then sure we can come up with a User class that represents a result from either query. In some of my widgets I defined auxiliary short-lived classes that make passing data around locally easier. It should be relatively easy to modify any Widget to have it take additional inputs when being constructed, but the key point is that this is done only on the classes that need it instead of having a global almighty omnipresent User class that could represent the result from one of dozens of queries.

lib/data/models/user/user_provider.dart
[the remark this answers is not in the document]
m-rosario: I could provide an alias for this and other types that get passed around a lot. How about AuthenticatedUser? I just don't want to obscure the fact that this is not a User as defined in the schema, and that it is instead the result of the getAuthenticatedUser query.

lib/schema/operations.graphql
m-rosario: Ah sure let me know if there are other conflicts I can resolve now

lib/data/models/user/user_provider.dart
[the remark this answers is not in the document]
andersan: That looks great to me. Thanks Miguel!

lib/data/models/user/user_provider.dart
```diff
@@ -1,24 +1,23 @@
...
 import 'package:shared_preferences/shared_preferences.dart';
-import 'package:uuid/uuid.dart';
 
-class UserProvider extends BaseProvider {
-  GraphQLClient client;
+import '../../../__generated/schema/operations.graphql.dart';
+import '../../../__generated/schema/schema.graphql.dart';
+import '../base/operation_result.dart';
+
+typedef AuthenticatedUser = Query$GetAuthenticatedUser$getAuthenticatedUser;
```
jonahtwalker, line 11: 🙏🏼

jonahtwalker (approved):
Thank you Miguel, and I appreciate everyone's thorough commentary! This is great!  🙌🏼

jupblb (approved):

lib/widgets/screens/messages/messages_list.dart
andersan: Thanks for addressing these questions and explaining the motivation here, Miguel. This sounds like a reasonable way to approach the different responses that will come from different GraphQL queries (indeed, even the class  `Query$GetAuthenticatedUser$getAuthenticatedUser` could have multiple different implementations depending on the amount of data which is fetched). 

The alias solution looks like an elegant way to handle this in the other case flagged by Michal.

andersan (approved):
Thanks for explaining the implications in the questions I had. This looks great and hopefully will result in a lot less boilerplate with all the graphql queries and mutations being generated. It'll be nice to be able to make small changes to `lib/schema/operations.graphql`  and `lib/schema/schema.graphql` rather than needing to add queries and mutations throughout the data/models/ directory.

lib/data/models/base/base_provider.dart
m-rosario: Thanks for suggesting this, I need to look into this a bit more to be able to give an informed opinion.

PR #41, opened 2023-06-29
Disable crashlytics for web

ashlilili:
Crashlytics doesn't support web. 
Disabling it for web so the app wouldn't crash when there's an error. Context: https://trello.com/c/Ji5gwDmI


jupblb:
It's a long lasting issue: https://github.com/firebase/firebase-js-sdk/issues/710

m-rosario (approved):
Thanks for cleanup up after me! Sorry if this caused you some trouble, I will look into the CrashHandler later to come up with a more definitive solution.

ashlilili:
> Thanks for cleanup up after me! Sorry if this caused you some trouble, I will look into the CrashHandler later to come up with a more definitive solution.

Thanks Miguel!

PR #43, opened 2023-06-29
Flutter web accessibility research doc

andersan:
This PR includes a document with some exploration of Flutter Web's a11y features. It makes some assumptions as to what a11y standards MicroMentor requires this app to meet - specifically, that we must be fully compliant with WCAG and Section 508.

The next steps here will occur outside the project:
1. Request additional expertise from Google on Flutter Web Accessibility specifically
2. Determine if we can just use the mobile apps to meet this requirement, instead of making sure we're 100% compliant on Flutter Web. (We'd still do some a11y work but the level of detail would be different.)


ashlilili:
Thanks a lot of writing this up Chris, this gives us a better understanding of the accessibility requirements and the questions we may have with flutter web development. I extracted the questions into the flutter support meeting doc for this Thursdays meeting and added some more followups, could you take a look when you get a chance and maybe add / change anything if you needed?

ashlilili (approved):
Thanks for writing this up Chris. This gives us a better understanding of the accessibility requirements and questions we may have with flutter web development. I extracted the questions into our Flutter meeting doc for this Thursday's meeting and added some followups. 
Could you take a look at it when you get a chance and add / change anything if needed?

PR #44, opened 2023-06-29
Home UI

ashlilili:
1. rename journey to progress
2. refactor upcoming session and resource section tiles to both use image section tile
3. changed some style and size

image

Simulator Screenshot - iPhone 14 Pro Max - 2023-06-29 at 18 54 40


lib/widgets/molecules/resources_section.dart
```diff
@@ -1,51 +1,57 @@
...
     //TODO(m-rosario): Implement callback for each tile.
     List<Widget> resourceTiles = [
-      Components.resourceListItemDivider,
-      ResourceTile(
-        icon: const Icon(Icons.school),
-        text: l10n.homeResourcesMentorTraining,
-        onPress: () => {},
+      ImageTile(
+        imageUrl:
+            "https://freevector-images.s3.amazonaws.com/uploads/vector/preview/40309/YOUNG_GIRL_WORKING_WITH_LAPTOP.jpg",
```
m-rosario, line 17: Generally, static images and similar content are stored in the source code as assets. This is a lot more efficient than downloading the same image every time this code is called. With avatars we use URL because these are stored externally, and could change at any time.

lib/constants/styles.dart
```diff
@@ -69,6 +69,14 @@ class ButtonStyles {
...
     );
   }
+
+  static ButtonStyle homePageButton(BuildContext context) {
+    return ButtonStyle(
+        shape: MaterialStateProperty.all<RoundedRectangleBorder>(
+            RoundedRectangleBorder(
+      borderRadius: BorderRadius.circular(Radii.buttonRadius),
+    )));
+  }
```
m-rosario, line 79: You can extend the ElevatedButton style instead of defining a ButtonStyle using the constructor. This is how it would look:

```
static ButtonStyle homePageButton(BuildContext context) {
    return ElevatedButton.styleFrom(
        shape: RoundedRectangleBorder(
      borderRadius: BorderRadius.circular(Radii.buttonRadius),
    ));
  }
```

In this case both look the same, but using this approach is generally better because it extends the ElevatedButton style instead of replacing it with a new one.

m-rosario (changes requested):
Thanks Ashley! Looks great, just left a couple of comments.

lib/widgets/molecules/resources_section.dart
ashlilili: Makes sense, changed to local assets

lib/constants/styles.dart
ashlilili: Gotcha, changed to extending ElevatedButton style

ashlilili:
> Thanks Ashley! Looks great, just left a couple of comments.

Thanks for the review! I addressed the comments in the newest commit

m-rosario (approved):
Good stuff

m-rosario (approved):

PR #45, opened 2023-06-30
Add a document about state management

jupblb:
As requested during standup, I'm providing a document summarizing my thoughts about employing a different state management framework in `mm-flutter-app`. TL;DR while Bloc looks great it would cost too much with too little benefit to migrate. This is of course open to discussion.


m-rosario (approved):
This very thorough and well written, thanks for looking into this! I agree that we should not put time and effort into rewriting the whole app unless we can find a solid reason why the current approach is not sufficient for our needs.

PR #46, opened 2023-07-03
Implement dashboard Invitations section with backend data

m-rosario:
- Updated mock server to generate inbox invitations. Also added the jobTitle field to the user generator, and made the avatarUrl field show a random image.
- Added UserProvider query for finding users by ID.
- Added ChannelsProvider query for getting inbox invitations.
- Updated Widgets to use backend data instead of hardcoded data.
- Pointed the 'See all' button to the Inbox page.

### DIFF
--- a/lib/schema/operations.graphql
+++ b/lib/schema/operations.graphql
@@ -15,7 +15,7 @@ query GetUserProfileInfo {
     }
 }
 
-query FindUsers ($filter: UserListFilter) {
+query FindAllUsers ($filter: UserListFilter) {
     findUsers(filter: $filter){
         id
         email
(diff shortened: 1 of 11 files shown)


lib/schema/operations.graphql
```diff
@@ -25,7 +25,39 @@ query FindUsers ($filter: UserListFilter) {
     }
 }
 
+query FindUsersWithFilter ($filter: UserListFilter) {
```
jupblb, line 28: Why do we have this separately to `FindAllUsers`? Isn't it better to add `jobTitle` as a field in `FindUsers`?

jupblb (commented):
Overall looks good but I'm not sure how I feel about merging PRs that base their functionality on mock server alone. While the reasons for that are understandable, I'm afraid this could potentially lead to being out of sync with the real backend.

I'd like @andersan to approve this PR before merging.

lib/schema/operations.graphql
m-rosario: These two queries have different purposes, even though they call the same endpoint. We do not need the jobTitle field for all users (not yet at least), so we instead query a list of users by ID using the filter function defined here. We could combine several queries into one query with all of the fields we could ever need, but then we lose the benefits of using GraphQL which is that we query exactly what we need when we need it.

m-rosario:
@jupblb I incorrectly said that the jobTitle field is not available in the backend. It is available under User, so the query that I created should work as is

andersan (approved):
The backend update looks good to me! It's possible that the one liner's implementation will change and we'll need to add a "shortDescription" or "oneLiner" field to the user. For now let's run with the job title.

PR #47, opened 2023-07-05
Make error logging and reporting optional for Queries and Mutations

m-rosario:
Logging/reporting can be disabled if the error is handled manually, as is the case of the StartScreen showing the Sign In page if it fails to get the authenticated user.


ashlilili (approved):
looks good!

PR #50, opened 2023-07-07
Complete GraphQL Codegen migration

m-rosario:
This removes all legacy models (User, Channel, Message) and uses the generated query result classes instead. Everything under data/models is now obsolete except for the providers, which are now under their own directory with a flatter and simpler hierarchy. Moving forward it should very easy to add a new query/mutation:

1.  Update the schema.graphql to include any new types or enums.
2. Update the corresponding operations.graphql file with the raw GQL query that will be used.
3. Run codegen to generate the corresponding Input and Output classes for the new query/mutation.
4. Add the query/mutation function to the corresponding provider, following the already established approach which is mostly boilerplate and should take less than 5 minutes.


lib/schema/operations_message.graphql
```diff
@@ -0,0 +1,70 @@
...
+        channelId
+        messageText
+        createdAt
+        replyToMessageId
+        deletedBy
+        updatedAt
+        deletedAt
+        editedAt
+        statuses {
+            seenAt
```
shadowbrush, line 46: You need the `userId` as well, otherwise you won't know who saw the message. Might not be critical on 1:1 conversations.

m-rosario: Added it, thanks!

lib/widgets/screens/home/users_list.dart
```diff
@@ -98,14 +100,17 @@ class UsersList extends StatelessWidget {
...
-        List users = data.response != null
+        List<Query$FindAllUsers$findUsers> users = data.response != null
             ? data.response!.reversed
                 .where((element) => element.id != currentUser?.id)
                 .toList()
             : [];
         users = users
             .where((element) =>
-                element.fullName.toLowerCase().contains(search.toLowerCase()))
+                element.fullName
```
andersan, line 110: We'll need to provide the query to do a paginated search (limit this to X results)

andersan (commented):
Found some time to go through this PR, and it all looks good to me. I'd like someone else to review this, too, since I'm not as familiar with the Flutter code base and may have missed something in these changes.

andersan (commented):
Found some time to go through this PR, and it all looks good to me. I'd like someone else to review this, too, since I'm not as familiar with the Flutter code base and may have missed something in these changes.

lib/widgets/atoms/reminder_banner.dart
```diff
@@ -91,7 +91,7 @@ class MaybeReminderBanner extends StatelessWidget {
       return const SizedBox(width: 0.0, height: 0.0);
     }, onData: (data, {refetch, fetchMore}) {
       int profileCompletionPercentage =
-          data.response!.profileCompletionPercentage;
+          data.response!.profileCompletionPercentage ?? 0;
```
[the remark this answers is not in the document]
m-rosario: Yes, because the schema defines this field as nullable, meaning that the generated model class also makes it nullable.

lib/widgets/screens/home/users_list.dart
m-rosario: Yes definitely! Here I was just refactoring the existing code to use the generated models. Pagination should be done separately.

andersan: Yes, this is something Holger is working on today in the backend

lib/providers/base/base_provider.dart
```diff
@@ -101,15 +103,17 @@ abstract class BaseProvider extends ChangeNotifier {
...
-    required DocumentNode document,
-    Map<String, dynamic>? variables,
-    bool logFailures = true,
-  }) async {
+  Future<QueryResult> runMutation(
+      {required DocumentNode document,
+      Map<String, dynamic>? variables,
+      bool logFailures = true,
+      FutureOr<void> Function(GraphQLDataProxy, QueryResult<Object?>?)?
+          update}) async {
```
jupblb, line 111: Nit: if you put a comma between `update` and `}` then it will be nicely formatted:
```dart
Future<QueryResult> runMutation({
  required DocumentNode document,
  Map<String, dynamic>? variables,
  bool logFailures = true,
  FutureOr<void> Function(GraphQLDataProxy, QueryResult<Object?>?)? update,
}) async {
```

jupblb (approved):
Sorry that the review took so long. It took more time than expected to set up new laptop. The changes look good to me (although I wish git diff was a bit smarter).

PR #51, opened 2023-07-11
UI for mentor/entrepreneur profile quick view card

m-rosario:
The widget parameters determine if the card displays the entrepreneur version or the mentor version. Note that the previous explore page was completely replaced with three hardcoded cards, just as a demonstration of how they would look. Once the Explore page frame is completed please replace the build Widget at explore.dart.

Some misc. additions in this PR:

1.  Added a blank-avatar.png asset that we can use as the default whenever the user doesn't have an avatar. The actual image can be replaced later, but we should be already implementing the null checking and defaulting to this asset.
2. Added an enums.dart to the constants library. Right now the only enum is UserType (mentor or entrepreneur). We will be doing a lot of checks for user type so it made sense to create this enum. Feel free to add more as needed.

image


ashlilili (dismissed):
Love it! This widget has so many small&customizable parts but you put them together in the code in such a clean way!

nit: would it be possible to have a colored outline around recommended cards like on the wireframe?

PR #54, opened 2023-07-12
Improvements to CrashHandler

m-rosario:
- Moved Firebase initialization to the main method of the app, since Firebase will be used for more than Crashlytics.
- Ensured error logging in logCrashReport is wrapped in a try-catch to prevent logger issues caused by the logger itself.
- Ensured Firebase Crashlytics is enabled only on platforms that support it (iOS and Android).
- Centralized crash reporting into a separate sendCrashReport method for better maintainability.
- Adjusted error classification to treat Error instances as fatal but without causing app exit.
- Removed _exitApp method, leaving app exit handling to the Flutter framework.
- Added Identifiers to the app_constants library for storing identifiers like the appName.
- Conducted overall refactoring for cleaner and more efficient code.

### DIFF
--- a/lib/constants/app_constants.dart
+++ b/lib/constants/app_constants.dart
@@ -5,6 +5,7 @@ import 'package:flutter/material.dart';
 part 'components.dart';
 part 'dimensions.dart';
 part 'enums.dart';
+part 'identifiers.dart';
 part 'paths.dart';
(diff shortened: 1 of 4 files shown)


ashlilili (approved):
LGTM. Thanks for improving the code health!

PR #55, opened 2023-07-12
Explore page vertical scroll, checkboxes and see more button

ashlilili:
Integrated the quick view card with the scroll. When see more button is pressed, 5 (configurable) more cards will be shown

Screenshot 2023-07-12 at 3 09 19 PM

Screenshot 2023-07-12 at 3 09 13 PM


package-lock.json
m-rosario, line 1: Was this file generated by accident?

lib/widgets/screens/explore/explore.dart
```diff
@@ -1,109 +1,227 @@
...
+    ]));
+  }
+}
+
+class ExploreScreen extends StatelessWidget {
+  const ExploreScreen({Key? key}) : super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
+    return const Scaffold(body: ExploreCardScroll());
```
m-rosario, line 225: It is generally not recommended to nest a Scaffold inside another Scaffold. The app already has a Scaffold widget in AppWrapper, which has the NavigationBar. In this particular case it seems you can just return the ExploreCardScroll directly and it will work just fine. If we want to use the features of the Scaffold in the child widgets then we have to do a bit of state management to be able to modify the one at the root level.

m-rosario (changes requested):
This looks great, very clever stuff! Just left a couple of comments no big deal. Thanks for taking care of this!

lib/widgets/screens/explore/explore.dart
ashlilili: Good catch! Removed this scaffold

package-lock.json
ashlilili: lol yup. removed!

m-rosario (approved):
Thanks!

PR #56, opened 2023-07-15
Explore page buttons

ashlilili:
Created the bottom buttons on the explore page.
1. Clear selected checkboxes
2. Send invites button, which is greyed out when no item is selected

https://github.com/micromentor-team/mm-flutter-app/assets/100324646/05eca5fc-ad1b-4238-8c9b-c7c63ad0eb8c


lib/constants/styles.dart
```diff
@@ -143,6 +143,20 @@ class TextStyles {
...
+    ThemeData theme = Theme.of(context);
+    return theme.textTheme.titleMedium?.copyWith(
+      color: theme.colorScheme.primary,
+    );
+  }
+
+  static TextStyle? bigButtonDisabledText(BuildContext context) {
+    ThemeData theme = Theme.of(context);
+    return theme.textTheme.titleMedium?.copyWith(
+      color: theme.colorScheme.surfaceVariant,
```
m-rosario, line 157: surfaceVariant is meant for surfaces not for text. In this particular case there is not enough contrast between the disabled button's color and the text color, so it is hard to read. You can try **onSurfaceVariant** for more readable text. 

In case you are curious about what these color names mean, the Material Design spec defines a set of color roles for surfaces and key elements, then there is a second set of colors that are meant to contrast with the first set when put on top (hence the "on" prefix). You can read more about color roles here if you are interested: https://m3.material.io/styles/color/the-color-system/color-roles

m-rosario (changes requested):
This is awesome! Just one comment regarding text readability.

lib/constants/styles.dart
ashlilili: Great read! Thanks for pointing this out. Changed to using onSurfaceVariant

m-rosario (approved):
🔥

PR #59, opened 2023-07-19
The UI for explore page filter section

rupalg:
This PR has the UI for the explore page filter section. This PR will be succeeded by multiple PRs that help mentors and entrepreneurs to filter their search. 
Screenshot 2023-07-19 at 6 37 39 PM

What's coming up next:
- the Search Filter page that shows the filters for the user
- The Advanced Filter page that shows the advanced filter options for the user
- The distinction between the mentor page and the entrepreneur page


m-rosario (changes requested):
The widget looks great visually, but it is currently not customizable. The requirements need this widget to display the selected filters, so your implementation should at the very least have a way to set the displayed text for each filter category: Expertise, Language, Location.

This is a good start, but it needs a bit more work to make it easier to integrate with the search page that Michal is working with.

EDIT: Ah I see you pushed a commit with exactly these changes before I could finish my review. Sorry about that, and thanks! Will take another look now.

m-rosario:
You addressed my review comment before I could even submit it! I will take another look

lib/widgets/atoms/explore_filter.dart
```diff
@@ -0,0 +1,150 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_gen/gen_l10n/app_localizations.dart';
+import '../../constants/app_constants.dart';
+
+class ExploreFilter extends StatelessWidget {
+  final UserType userType;
+  final bool userFiltersSelected;
+  final List<String?>? skills;
+  final List<String?>? languages;
+  final List<String?>? countries;
```
m-rosario, line 10: nit: You could assume that if any of the filter lists is not empty then userFiltersSelected will be true, having to set both the filters and the flag is a bit redundant. You could have the flag as a private instance field that is initialized based on whether any of the lists (skills, languages, countries) is not empty. 

Not a huge deal though, just a small optimization :-)

m-rosario (approved):
Looks great, thanks Rupal!

m-rosario:
Also flutter-analyze seems to be failing because of redundant null check operators (!) in this part of the code:

```
  String joinFirstThree(List<String?>? strings) {
    int count = strings!.length < 3 ? strings!.length : 3;
    List<String> firstThreeElements = [];
    for (int i = 0; i < count; i++) {
      firstThreeElements.add(strings![i]!);
    }
    return firstThreeElements.join(' • ');
  }
```
Using `strings!` once guarantees that the `strings` variable will not be null (or else app will crash). Therefore there is no need to check it a second and third time.

rupalg:
Hi Miguel, 
Thanks for pointing out the error with the code earlier and suggesting the optimization with the "userFiltersSelected" variable. I have made the needed changes. 

Again, I appreciate your quick reviews! 👯

lib/widgets/atoms/explore_filter.dart
```diff
@@ -1,18 +1,20 @@
+import 'dart:js_interop';
+
 import 'package:flutter/material.dart';
 import 'package:flutter_gen/gen_l10n/app_localizations.dart';
 import '../../constants/app_constants.dart';
 
 class ExploreFilter extends StatelessWidget {
   final UserType userType;
-  final bool userFiltersSelected;
+  // final bool userFiltersSelected;
```
m-rosario, line 7: If this is not needed then it is best to delete it to avoid cluttering up the code.

lib/widgets/atoms/explore_filter.dart
```diff
@@ -79,6 +81,8 @@ class ExploreFilter extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final AppLocalizations l10n = AppLocalizations.of(context)!;
+    bool userFiltersSelected =
+        (skills!.isNotEmpty || languages!.isNotEmpty || countries!.isNotEmpty);
```
m-rosario, line 83: You need to handle the case when the lists are null or else the app will crash right here. You can either make the lists a required field (caller passes an empty list if no filters are applied), or keep them as nullable (preferred) and do null checking on the lists before using them. 

There are other operators besides the dangerous null-assertion operator (!): you could use the conditional access operator (?), the null-coalescing operator (??), or plain old 'if' operators to check equality against 'null'.

This codelab has some useful tips: https://dart.dev/codelabs/null-safety

m-rosario (changes requested):
Added a few comments!

lib/widgets/atoms/explore_filter.dart
rupalg: Absolutely, removed.

lib/widgets/atoms/explore_filter.dart
```diff
@@ -1,18 +1,20 @@
...
 import 'package:flutter_gen/gen_l10n/app_localizations.dart';
 import '../../constants/app_constants.dart';
 
 class ExploreFilter extends StatelessWidget {
   final UserType userType;
-  final bool userFiltersSelected;
+  // final bool userFiltersSelected;
   final List<String?>? skills;
   final List<String?>? languages;
   final List<String?>? countries;
```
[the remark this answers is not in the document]
rupalg: Good call, I can make it List<String>? instead

lib/widgets/atoms/explore_filter.dart
rupalg: Hi Miguel, I went through the codelab on null safety. It's a great resource. Thanks for sharing. 
Yeah, I agree that it is preferred to keep the lists nullable, in the scenario there are no filters applied. I have added some more logic that takes away the need for the null-assertion using some if statements. Please take a look.

lib/widgets/atoms/explore_filter.dart
```diff
@@ -67,7 +69,7 @@ class ExploreFilter extends StatelessWidget {
     );
   }
 
-  String joinFirstThree(List<String?>? strings, AppLocalizations l10n) {
+  String joinFirstThree(List<String?>? strings) {
     int count = strings!.length < 3 ? strings.length : 3;
     List<String> firstThreeElements = [];
     for (int i = 0; i < count; i++) {
```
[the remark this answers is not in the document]
rupalg: Wow, that is quite sleek. I'll use this syntax.

rupalg:
@m-rosario addressed the requested changes :)

PR #60, opened 2023-07-19
Implement storybook_flutter, to review UI components independent of the business logic

jonahtwalker:
Stands up [storybook_flutter](https://pub.dev/packages/storybook_flutter) and two refactored MockProviders to support this. In `storybook.dart`, the `main` function can be run to spin up the UI showcase. Here, the large majority of widgets are covered and available to be interacted with. This was a great exercise to get back up to speed with the Flutter side of things, but there are a few bridges left to cross that will see updates as I work in other places in the source.

Some challenges:
- RTL is not officially supported. There's some documentation on extending the functionality to support "global" knobs, but this will be a phase 2 exercise when I've got some more time to work on this.
- Global knobs: There are a few already (dark/light, device and orientation) but there should be a few more like `theme` and `localization`
- Passing the providers to permeate each level of the widget. I'm having trouble with partially rendering most screen widgets, in that the frame is there but the content is not able to access the provider and throws an error.

To do in next phase:
- [ ] Solve the provider issues for the screen stories
- [ ] Mock Messages provider
- [ ] Implement routing
- [ ] Implement more global knobs
- [ ] Host this independently so it's accessible to the _rest_ of the Team (?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,15 +116,6 @@ In Android Studio, select a target platform (Chrome web browser, Android
 Simulator, etc.) and then click on the `Run main.dart` button to build and the
 run the app on that platform. Shortcut : `Control + R` (macOS).
 
-### Run Widgetbook
-
-In Android Studio, select `Edit Configurations` in the Run menu. Copy
-configurations of main.dart and edit name as `widgetbook`. Edit
-`Dart entrypoint` to the path of your Widgetbook's main.dart. For e.g.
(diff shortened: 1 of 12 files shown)


m-rosario:
This is very cool! However it looks like it will take a considerable amount of effort to implement and then maintain this. I am specially worried about the mock provider functions. Do you think that instead of overriding the functionality of the app we could use the already existing mock server?

PR #64, opened 2023-07-25
Invite-to-connect page

ashlilili:
https://github.com/micromentor-team/mm-flutter-app/assets/100324646/bd1553a7-d314-4c05-bab4-f140e76f5b63

---
1. Invite to connect page: sends connect invites to one to multiple users. When multiple users are selected, show a scroll of user names and profile image. When only one user is selected, show a quick view card.
2. Refactor: move ProfileQuickViewInfo and helper functions to profile_quick_view_card.dart to make it reusable across files

### DIFF
--- a/lib/constants/dimensions.dart
+++ b/lib/constants/dimensions.dart
@@ -37,7 +37,7 @@ class Dimensions {
   static const double avatarRoundedRectLength = 28.0;
   static const double drawerHeaderHeight = 64.0;
   static const double minimumInteractionWidgetLength = 48.0;
-  static const Size imageTile = Size(88.0, 160.0);
+  static const Size imageTile = Size(88.0, 168.0);
   static const Size imageTileRectangularImage = Size(88.0, 72.0);
   static const double reminderBannerImageHeight = 80.0;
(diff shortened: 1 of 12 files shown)


m-rosario (approved):
Thanks Ashley this is great!

In case you are wondering, the reason your ClosableTile onSurface border looks much darker than the one in the Figma design is because the one in Figma is incorrectly referencing onSurface from the DARK theme, instead of the Light theme that we are using. The design tokens approach doesn't work unless we all use color roles correctly... sorry about that! I will point this out to UX and we can revise this later.

ashlilili:
> Thanks Ashley this is great!
> 
> In case you are wondering, the reason your ClosableTile onSurface border looks much darker than the one in the Figma design is because the one in Figma is incorrectly referencing onSurface from the DARK theme, instead of the Light theme that we are using. The design tokens approach doesn't work unless we all use color roles correctly... sorry about that! I imagine the color role they were looking for is 'outline'. I will point this out to UX and we can revise this later.

I see thanks for pointing this out and thank you for the quick review!

PR #65, opened 2023-07-26
Add name parameter to the routes

LucasMagnum:
To reduce the synchronisation effort between the backend and frontend, I suggest having names for all of our routes.  

That will separate the route/path definition inside the flutter application and the route name used externally (for push notifications for instance). 

The idea here is that we can modify the internal pathing freely as long as we do not change the name of the route. Of course, in case of new parameters added to the route we would also need to change it in the backend. 

Hence, I have changed all the routes to contain a name and updated the references to use the "Routes.<route>.path".

### DIFF
--- a/lib/constants/routes.dart
+++ b/lib/constants/routes.dart
@@ -1,24 +1,38 @@
 part of 'app_constants.dart';
 
+class Route {
+  final String name;
+  final String path;
+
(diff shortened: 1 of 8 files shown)


m-rosario (approved):
Thanks!

PR #69, opened 2023-07-29
Update mock server to generate a conversation with other user

m-rosario:
Add a mock conversation so that the Chat UI can be tested more easily. The mock conversation covers the following:

- Conversation spanning multiple days
- Messages with emoji only
- Messages with text and emoji
- Replies from both sides
- Deleted message on both sides
- Edited message
- 2 unseen messages
- Allow adding new messages that persist in memory while the mock server is running

Includes some light refactoring of existing code.


andersan (approved):
Looks great, thanks Miguel!

PR #72, opened 2023-07-31
Clean up app Dimensions constants

m-rosario:
Removed measurements constants that were specific to a single component, and therefore were not being reference elsewhere.

The following Insets were renamed to follow the common UI measurement naming convention of T-Shirt sizes:
widgetSmallestInset  -> paddingExtraSmall
widgetSmallInset -> paddingSmall
widgetMediumInset -> paddingMedium
widgetMediumLargeInset -> paddingLarge
widgetLargeInset -> paddingExtraLarge

Also added some constants for window classes, which will become useful as we move towards designing adaptive layouts.

### DIFF
--- a/lib/constants/dimensions.dart
+++ b/lib/constants/dimensions.dart
@@ -1,26 +1,43 @@
 part of 'app_constants.dart';
 
+class LayoutBreakpoints {
+  const LayoutBreakpoints._private();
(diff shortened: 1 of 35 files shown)


ashlilili (approved):
lgtm, thanks for taking care of this

PR #73, opened 2023-07-31
GraphQL Schema updates with options/content

andersan:
This PR contains the following edits:

- Updated graphql schema to include some random updates from last few weeks, including pagination, sorting, and other options. See this PR in the backend for more context on those options: https://github.com/micromentor-team/mmdata/pull/239 
- Updates to mock server to keep up with the above schema changes, already reviewed in #84
- Added content provider, which includes static options like Countries, Languages, Company Stages, Company Types, Expertises, Industries, some preset options for Genders, and "Education Levels". Later, it may provide additional options. Countries may be broken out into a separate Location provider once we determine how States/Cities are going to work.
- Added 2 queries to the User provider which return fields specific to Mentors or Mentees. Those queries still need appropriate filters if they are to be expected to return only mentor or only mentee users.

Example for how to fetch mentees:
```dart
final menteeUsers = userProvider.findMenteeUsers(
  optionsInput: Input$FindObjectsOptions(
    limit: 10, // only return 10 users
    skip:
      10, // return users after the first 10 (e.g. for the 2nd page of a search)
    sort: [
      Input$SortItem(
        direction: Enum$SortDirection.desc, field: "createdAt")
    ], // sort by createdAt descending
  ),
  filterInput: Input$UserListFilter(
    ids: null,
    searchText:
      "" // some non-empty field seems to be required by our user search at the moment, buggy but works
    ),
  matchInput: Input$UserInput(seeksHelp: true // mentee users
    ));

menteeUsers.then((value) => debugPrint(
  // log all attributes from each user
  value.response!.map((e) => e!.toJson()).toList().toString(),
));
```

And to fetch countries:
```dart
final contentProvider = Provider.of<ContentProvider>(context);
final content = contentProvider.findCountries();

// log content
content.then((value) => debugPrint(
    value.response!.map((e) => e!.toJson()).toList().toString()));
```

Unless only one of the types is needed from the Content provider, it's probably better to use `findAllOptionsByType()` rather than the individual queries like `findCountries`.

<details>
  <summary>Hidden graphql for the above dart code for mentees</summary>
  
 ```graphql
query FindMenteeUsers($options: FindObjectsOptions, $filter: UserListFilter, $match: UserInput) {
    findUsers(options: $options, filter: $filter, match: $match) {
        id
        email
        firstName
        lastName
        fullName
        avatarUrl
        userHandle
        cityOfResidence
        regionOfResidence
        countryOfResidence {
            translatedValue
        }
        jobTitle
        groupMemberships {
            ... on MenteesGroupMembership {
                soughtExpertises {
                    translatedValue
                }
                industry {
                    translatedValue
                }
            }
        }
        companies {
            companyStage {
                translatedValue
            }
        }
    }
}
```

Variables JSON:
```json
{
  "filter": {
    "ids": null,
    "searchText": ""
  },
  "match": {
    "seeksHelp": true
  },
  "options": {
    "limit": 500,
    "skip": 0,
    "sort": [
      {
        "direction": "desc",
        "field": "createdAt"
      }
    ]
  }
}
```
</details>

### DIFF
--- a/lib/providers/content_provider.dart
+++ b/lib/providers/content_provider.dart
@@ -1,12 +1,138 @@
-// import 'package:flutter/material.dart';
-// import 'package:graphql_flutter/graphql_flutter.dart';
-// import 'package:mm_flutter_app/__generated/schema/operations_content.graphql.dart';
-// import 'package:mm_flutter_app/__generated/schema/schema.graphql.dart';
+import 'package:flutter/material.dart';
+import 'package:graphql_flutter/graphql_flutter.dart';
+import 'package:mm_flutter_app/__generated/schema/operations_content.graphql.dart';
 
-// import 'base/base_provider.dart';
-// import 'base/operation_result.dart';
+import 'base/base_provider.dart';
+import 'base/operation_result.dart';
 
-// class ContentProvider extends BaseProvider {
-//   ContentProvider({required super.client});
-//   // TODO: add queries for expertises, education levels, countries, company stages, company types, genders, industries, and languages
-// }
+typedef CompanyStage = Query$FindCompanyStages$findCompanyStages;
+typedef CompanyType = Query$FindCompanyTypes$findCompanyTypes;
+typedef Country = Query$FindCountries$findCountries;
+typedef EducationLevel = Query$FindEducationLevels$findEducationLevels;
+typedef Expertise = Query$FindExpertises$findExpertises;
+typedef Gender = Query$FindGenders$findGenders;
+typedef Industry = Query$FindIndustries$findIndustries;
+typedef Language = Query$FindLanguages$findLanguages;
+
+class ContentProvider extends BaseProvider {
+  List<CompanyStage>? _companyStages;
+  List<CompanyType>? _companyTypes;
+  List<Country>? _countries;
+  List<EducationLevel>? _educationLevels;
+  List<Expertise>? _expertises;
+  List<Gender>? _presetGenders;
+  List<Industry>? _industries;
+  List<Language>? _languages;
+
+  ContentProvider({required super.client});
+
+  void _setAllContentOptions({
+    List<CompanyStage>? companyStages,
+    List<CompanyType>? companyTypes,
+    List<Country>? countries,
+    List<EducationLevel>? educationLevels,
+    List<Expertise>? expertises,
+    List<Gender>? presetGenders,
+    List<Industry>? industries,
+    List<Language>? languages,
+  }) {
+    if (companyStages != null) {
+      _companyStages = companyStages;
+    }
+    if (companyTypes != null) {
+      _companyTypes = companyTypes;
+    }
+    if (countries != null) {
+      _countries = countries;
+    }
+    if (educationLevels != null) {
+      _educationLevels = educationLevels;
+    }
+    if (expertises != null) {
+      _expertises = expertises;
+    }
+    if (presetGenders != null) {
+      _presetGenders = presetGenders;
+    }
+    if (industries != null) {
+      _industries = industries;
+    }
+    if (languages != null) {
+      _languages = languages;
+    }
+    debugPrint('Updated content provider values: ${toString()}');
+  }
+
+  // getters
+
+  List<CompanyStage>? get companyStageOptions {
+    return _companyStages;
+  }
+
+  List<CompanyType>? get companyTypeOptions {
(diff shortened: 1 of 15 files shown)


m-rosario:
@andersan
Added a commit to regenerate the GraphQL Codegen classes. If you make more changes to any of the files inside the _schema_ directory you can run `dart run build_runner build` from the root of the project to regenerate as needed.

The only change to the code that I could notice is that now profileCompletionPercentage doesn't need any null checks, since it is now a required field. I updated accordingly. Thanks!

andersan:
Thanks, Miguel! I forgot about the need to run that graphql code generator after updating the schema.

I plan to provide more queries, including the following localized/translated lists of content:
- expertises
- industries
- company stages
- company types
- languages (some can be used as UI/display languages, others are only used for matching users)

The search queries with the GraphQL `UserSearch` instead of `findUsers` probably need more work before we use them in the explore screen. 

Many of the fields included in the mockup on the profile screen will be implemented later (after this PR, maybe next week). Maybe we'll wait until the onboarding flow is defined in the PRD and mockups before we start adding to the backend.

andersan:
@m-rosario and @jupblb , I started adding to a new Content provider which will supply options like Countries, Expertises, Industries and Languages from the backend.

Could you let me know if you think this is a good approach? If so, I'll add the remaining types. Right now the proivder is only implemented with Countries as a model, and I haven't checked to make the app still works properly yet.

m-rosario:
> @m-rosario and @jupblb , I started adding to a new Content provider which will supply options like Countries, Expertises, Industries and Languages from the backend.
> 
> Could you let me know if you think this is a good approach? If so, I'll add the remaining types. Right now the proivder is only implemented with Countries as a model, and I haven't checked to make the app still works properly yet.

@andersan This looks good to me! Having the data stored in the instance fields and accessed through getters is precisely what I am looking to improve on the other providers. I also look to have the queries return a Future<OperationResult> instead of a Widget. Callers should be able to request data either from the query or from the getters and be responsible for building their own Widgets. But you can continue doing the rest like this and I can take care of refactoring once we have the optimal approach perfected. Thanks!

andersan:
@m-rosario Thanks for checking out these changes!

I see what you mean. `runQuery` from the base provider returns a widget, so this here does too. After you make the change to `BaseProvider` and all the other providers, I can implement the rest of the queries here. That way, we avoid the extra work of refactoring them.

andersan:
just rebased with main, hopefully more changes coming this afternoon after a day full of meetings

lib/providers/content_provider.dart
```diff
@@ -0,0 +1,352 @@
...
+          ? Query$FindAllOptionsByType.fromJson(
+              queryResult.data!,
+            )
+          : null,
+    );
+    if (result.response != null) {
+      _setAllContentOptions(
+        companyStages: result.response?.findCompanyStages
+            .map((e) =>
+                Query$FindCompanyStages$findCompanyStages.fromJson(e.toJson()))
```
andersan, line 321: This `fromJson` `toJson` feels wrong but I didn't know how else to assign a variable of type  `Query$FindAllOptionsByType$findIndustries` to `Query$FindIndustries$findIndustries`.

Maybe a bit of premature optimization here trying to do this all in one request rather than the 10+ requests to do everything separately, as we will probably only do this once on the app.

m-rosario (approved):
Looks great, thanks @andersan !

I just have a question. Why do most of the queries in the frontend take no parameters, even though the schema allows providing a UiLanguage enum? Is this being inferred from somewhere else?

andersan:
Hey Miguel, 

Thanks for checking this out.

The language is assigned automatically for users based on the authorization headers for logged-in users, or the `AcceptLanguage` header. That is how we envision the Flutter app to communicate with this, but if we need to use that UILanguage enum in the GraphQL, that works too.

When we first implemented this feature, that parameter wasn't there, but we added it since we think it may be useful for potential API integrations with clients. It makes the language field more explicit. Still not 100% sure it's needed due to `AcceptLanguage` but it's there in case we do.

The logic for handling the languages automatically is [described here](https://github.com/baragaun/secureid-service/blob/main/src/services/secureId/docs/handling-user-languages.md). In short, for a logged-in user, it refers to a new field `preferredUiLanguage`, falls back to the device language if we support that and the `preferred` language isn't supported, and falls back to English if that's not supported.

PR #74, opened 2023-08-01
Query backend to get the last message of each channel

m-rosario:
This is used to build the preview in the list of channels. The query for unseenMessages is removed and combined with the query for all messages to avoid making two separate queries on the same data.

### DIFF
--- a/lib/widgets/screens/inbox/inbox_chats.dart
+++ b/lib/widgets/screens/inbox/inbox_chats.dart
@@ -1,5 +1,6 @@
 import 'package:flutter/material.dart';
 import 'package:go_router/go_router.dart';
+import 'package:mm_flutter_app/__generated/schema/schema.graphql.dart';
 import 'package:mm_flutter_app/constants/app_constants.dart';
 import 'package:mm_flutter_app/utilities/debug_logger.dart';
 import 'package:mm_flutter_app/utilities/router.dart';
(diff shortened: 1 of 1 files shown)


lib/widgets/screens/inbox/inbox_chats.dart
```diff
@@ -84,15 +70,40 @@ class _InboxChatsScreenState extends State<InboxChatsScreen>
...
-            avatarUrl: channelAvatarUrl,
-            fullName: channelName,
-            date: DateTime.now().subtract(Duration(
-                days: i)), // TODO(m-rosario): Use date from last message
-            message: '', // TODO(m-rosario): Get text from last message
-            notifications: channelUnseenMessages.length,
-            onPressed: () =>
-                context.push('${Routes.inboxChats.path}/${channel.id}'),
+          child: messagesProvider.findChannelMessages(
+            input: Input$ChannelMessageListFilter(channelId: channel.id),
```
andersan, line 74: Sent this to @m-rosario in chat but moving discussion here for visibility. 

This request should get only the text from the last message and the channel ID. 

```graphql
query FindChannelMessages($filter: ChannelMessageListFilter, $options: FindObjectsOptions) {
  findChannelMessages(filter: $filter, options: $options) {
    channelId
    messageText
  }
}
```
```json
{
  "filter": {
    "channelId": "647df8b494e2b5c62724c276"
  },
  "options": {
    "limit": 1,
    "sort": [
      {
        "direction": "desc",
        "field": "createdAt"
      }
    ]
  }
}
```

m-rosario: @andersan I tried to implement this, but the schema defines the 'sort' field as a String, meaning that I cannot provide the options that you're giving me here. Is there a newer version of the schema?

andersan: @m-rosario yes, the newer version is in #73. Didn't realize that the schema had changed here, Holger implemented this a few weeks ago.

I could merge #73 into main and finish off the content provider later if that'd be helpful.

m-rosario: @andersan Don't worry about it I will just update the query once it's merged. I was just curious why it didn't work

PR #75, opened 2023-08-01
First hardcoded iteration of the profile page scaffolding and profile page bar

rupalg:
Hello all,

To unblock folks working on the profile page, I have created the blank page along hardcoded profile page bar on the top [UI only].

I will make this profile page top bar functional as well in future commits. This is what it looks like at the moment:

Screenshot 2023-08-01 at 2 48 00 PM

I also want to give a shoutout to the UX team- Katy and Ujjaini for adding the necessary specs for the UI elements on the Figma file.


LucasMagnum (approved):
LGTM!

PR #76, opened 2023-08-01
Create mock resolvers for replying, deleting, and updating messages

m-rosario:
This allows deleting/undeleting, editing messages, creating replies, and marking messages as seen in the mock server. These changes persist as long as the server is running.

I need confirmation on the logic for the backend. I know these fields are all that is needed for the frontend, but I would like to know if there are side effects to running these operations that we need to be aware of. For example, does editing a message update the `editedAt` and `updatedAt` fields? Does deleting or undeleting a message update the `deletedAt` and `deletedBy` fields?

Also, is the backend able to tell the difference between a `deletedAt` field that is null because it wasn't set by the caller in the `updateChannelMessage` mutation, or because it was specifically set to null by the caller (to undelete the message)?

### DIFF
--- a/mm-mock-server/src/mocks/mutations.ts
+++ b/mm-mock-server/src/mocks/mutations.ts
@@ -31,6 +31,14 @@ export function mockMutations(serverState: MockServerState) {
             serverState.channelMessages.push(message);
             return message;
         },
+        deleteChannelMessage: (_: any, args: { channelMessageId: string, deletePhysically: boolean }) => {
+            var channelMessage = serverState.channelMessages.find((element) => element.id == args.channelMessageId);
+            const currentTime : string = new Date().toISOString();
+            channelMessage.deletedAt = currentTime;
+            channelMessage.updatedAt = currentTime;
+            channelMessage.deletedBy = channelMessage.createdBy;
+            return "ok";
+        },
         updateChannelInvitation: (_: any, args: { input: { status: string }}) => {
             const invitation = serverState.channelInvitations[0]
             if (args.input.status) {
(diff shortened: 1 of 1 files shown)


andersan:
Hey Miguel, I'll look into this this morning and get back to you. I'm not sure on the details of the message deletion fields (`deletedAt`) and their behavior on being deleted or "undeleted", or whether the "un-deletion" flow has been considered at all when setting this up.

> does editing a message update the editedAt and updatedAt fields?
It should update the `updatedAt` field whenever any changes are made, and `editedAt` is updated only if the message text changes.

andersan:
>Does deleting or undeleting a message update the `deletedAt` and `deletedBy` fields?

Deleting via the graphql `deleteChannelMessage` mutation does update `deletedAt`, but not `deletedBy`. `deletedBy` will always be undefined unless it's set via `updateChannelMessage`. We could implement that so that it is set whenever objects are deleted, which we haven't done yet. I think this is the first use case where any data needs to be recovered after deletion.

There isn't an explicit `undelete` functionality built for messages, so we'll need to add that. I think that it would work to update a deleted message to set `deletedAt` to null, but I don't know for sure. It's not clear to me what the implications here are: do we expect this message to still be returned on the `Channel.messages` GraphQL field after deletion? If so, we may need to make some other changes in the backend to make sure it's still returned. I'm assuming we wouldn't want the message text to be returned to the frontend, as the message is "deleted", so that'd need to be implemented too (return the message ID, createdAt, but not messageText).

> is the backend able to tell the difference between a `deletedAt` field that is null because it wasn't set by the caller in the updateChannelMessage mutation, or because it was specifically set to null by the caller (to undelete the message)?

No, it doesn't know the difference. 

Some other context: `deletePhysically` should be `false` if we want to be able to undelete something. If it is undefined or true, the data will be permanently deleted from MongoDB.

andersan (dismissed):

andersan:
It sounds like the front-end will still be downloading deleted messages for the recipient, but not displaying that data to the user.

In practice, this means that the deleted message is still served to both users and downloaded to their devices, right? I don't think this is something we need to address immediately but could add to the backlog (and in any case, isn't part of this PR).

andersan (approved):
Changes to the mock resolvers all look good! Thanks for asking about how message deletion is implemented on our backend.

PR #77, opened 2023-08-02
Route user to the right path after receiving a notifications - WIP

LucasMagnum:


LucasMagnum:
Just for the context, this is the expected behavior for now. The foreground will be fixed when we use sometehing like https://pub.dev/packages/flutter_local_notifications, that will be covered later when the push notifications cames back to priority.

LucasMagnum:
I will sync and merge this code after: https://github.com/micromentor-team/mm-flutter-app/pull/97#pullrequestreview-1575420792

PR #80, opened 2023-08-03
Add experience and education info to profile overview

jupblb:
The underlying data is hardcoded because the `User` model:
1) Doesn't have education data
2) Contains a `Company` structure but it seems to be related only to founded busiesses

### DIFF
--- a/lib/widgets/molecules/profile_experience_and_education.dart
+++ b/lib/widgets/molecules/profile_experience_and_education.dart
@@ -1,4 +1,6 @@
 import 'package:flutter/material.dart';
+import 'package:flutter_gen/gen_l10n/app_localizations.dart';
+import 'package:url_launcher/url_launcher.dart';
 import 'package:mm_flutter_app/constants/app_constants.dart';
 
 class ExperienceAndEducation extends StatelessWidget {
@@ -39,6 +41,7 @@ class _Experience extends StatelessWidget {
       start: DateTime.utc(2016, 3),
       end: DateTime.utc(2019, 5),
       location: "Cleveland, Ohio, USA",
+      companyUrl: "https://mcconsulting.com/",
     ),
   ];
 
@@ -47,21 +50,37 @@ class _Experience extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final theme = Theme.of(context);
+    final l10n = AppLocalizations.of(context)!;
     final items = _exampleExperience
-        .map((exp) => [
-              Text(exp.position, style: theme.textTheme.labelMedium),
-              Text(exp.companyName),
-              Text("(${exp.dateRange}) · ${exp.timeRange}"),
-              Text(exp.location),
-              const SizedBox(height: Insets.paddingMedium),
-            ])
+        .map((exp) {
+          final companyName = (exp.companyUrl == null)
+              ? Text(exp.companyName)
+              // https://stackoverflow.com/a/54540261
+              : InkWell(
+                  onTap: () => launchUrl(Uri.parse(exp.companyUrl ?? "")),
+                  child: Text(
+                    exp.companyName,
+                    style: const TextStyle(
+                        decoration: TextDecoration.underline,
+                        color: Colors.blue),
+                  ),
+                );
+
+          return [
+            Text(exp.position, style: theme.textTheme.labelMedium),
+            companyName,
+            Text("(${exp.dateRange(l10n.present)}) · ${exp.timeRange}"),
+            Text(exp.location),
+            const SizedBox(height: Insets.paddingMedium),
+          ];
+        })
         .expand((i) => i)
         .toList();
 
     return Column(
       crossAxisAlignment: CrossAxisAlignment.start,
       children: [
-        Text("My experience", style: theme.textTheme.labelLarge),
+        Text(l10n.profileExperience, style: theme.textTheme.titleMedium),
         const SizedBox(height: Insets.paddingMedium),
         ...items,
       ],
@@ -98,6 +117,7 @@ class _Education extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final theme = Theme.of(context);
+    final l10n = AppLocalizations.of(context)!;
     final items = _exampleEducation
         .map((edu) => [
               Row(children: [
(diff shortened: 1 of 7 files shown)


lib/widgets/molecules/profile_experience_and_education.dart
```diff
@@ -0,0 +1,180 @@
...
+              Text(exp.location),
+              const SizedBox(height: Insets.paddingMedium),
+            ])
+        .expand((i) => i)
+        .toList();
+
+    return Column(
+      crossAxisAlignment: CrossAxisAlignment.start,
+      children: [
+        Text("My experience", style: theme.textTheme.labelLarge),
```
m-rosario, line 64: 1. Please create a l10n variable for this string
2. The labelLarge style is too similar to the one being used inside each section, so it is hard to tell where one section ends and the other begins. The Figma mock uses titleMedium.

The same two thing apply to the 'My education' section.

lib/widgets/molecules/profile_experience_and_education.dart
```diff
@@ -0,0 +1,180 @@
...
+      end: DateTime.utc(2021, 11),
+      location: "Cleveland, Ohio, USA",
+    ),
+    _ExperienceItem(
+      position: "Fonder",
+      companyName: "MC Consulting",
+      start: DateTime.utc(2016, 3),
+      end: DateTime.utc(2019, 5),
+      location: "Cleveland, Ohio, USA",
+    ),
```
m-rosario, line 42: The mock and the PRD mention an optional website link. Can you add this field and make the url 'clickable'?

lib/widgets/molecules/profile_experience_and_education.dart
```diff
@@ -0,0 +1,180 @@
...
+
+  const _ExperienceItem({
+    required this.position,
+    required this.companyName,
+    required this.start,
+    this.end,
+    required this.location,
+  });
+
+  String get dateRange => "${start.year} - ${end?.year ?? "present"}";
```
m-rosario, line 140: Use a localized variable for "present"

lib/widgets/molecules/profile_experience_and_education.dart
```diff
@@ -0,0 +1,180 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    final theme = Theme.of(context);
+    final items = _exampleExperience
+        .map((exp) => [
+              Text(exp.position, style: theme.textTheme.labelMedium),
+              Text(exp.companyName),
+              Text("(${exp.dateRange}) · ${exp.timeRange}"),
+              Text(exp.location),
```
m-rosario, line 55: In this entire widget, everything that is not a titleMedium ('My experience' and 'My education') uses textTheme.bodySmall.

The body text that is bold uses fontWeight 700, like this:

```
Text(
    exp.position,
    style: theme.textTheme.bodySmall?.copyWith(
       fontWeight: FontWeight.w700,
    ),
),
```

m-rosario (changes requested):
Left some comments, mostly style related. Good stuff!

lib/widgets/molecules/profile_experience_and_education.dart
jupblb: Is... this a problem?

lib/widgets/molecules/profile_experience_and_education.dart
jupblb: Done. This makes sense. But I wonder - how did you notice that Figma mock uses `titleMedium`?

lib/widgets/molecules/profile_experience_and_education.dart
jupblb: Fixed

lib/widgets/molecules/profile_experience_and_education.dart
m-rosario: In the Figma mock you can check the properties of all components. I made a post about  this on our the team's chat on July 18. Here is an example:

image

lib/widgets/molecules/profile_experience_and_education.dart
m-rosario: Each typescale is optimized for a different use case, body works better for longer lines of text. Also this is what UX used in the design, and we agreed to follow it as closely as possible. It is not a problem per-se and the change is barely noticeable, I am just letting you know that you can get these styles from the mock very easily instead of trying to approximate them.

lib/widgets/molecules/profile_experience_and_education.dart
```diff
@@ -47,21 +50,37 @@ class _Experience extends StatelessWidget {
...
+              // https://stackoverflow.com/a/54540261
+              : InkWell(
+                  onTap: () => launchUrl(Uri.parse(exp.companyUrl ?? "")),
+                  child: Text(
+                    exp.companyName,
+                    style: const TextStyle(
+                        decoration: TextDecoration.underline,
+                        color: Colors.blue),
+                  ),
+                );
```
m-rosario, line 67: The design has the website as a separate line at the bottom instead of turning the company into a hyperlink.

m-rosario (changes requested):

lib/widgets/molecules/profile_experience_and_education.dart
m-rosario: Also, note that the text using the label text style is notably smaller than the rest of the text in the body. If you use the same text style (body) for all of it and then increase the fontWeight for the bold part it will look more consistent.

Sorry if these seem like nits, but this is literally what UI is all about.

lib/widgets/molecules/profile_experience_and_education.dart
jupblb: Thanks! I've adjusted the text style everywhere. Will remember in the future.

lib/widgets/molecules/profile_experience_and_education.dart
jupblb: Changed. I used `FontWeight.bold` as it's the same as `.w700`.

lib/widgets/molecules/profile_experience_and_education.dart
jupblb: Fixed.

m-rosario (approved):
Thanks Michal!

PR #82, opened 2023-08-03
iteration one of connecting recommended mentors with the mock server

rupalg:
Since the creation of the UI of the recommended mentor's list, this PR connects the recommended mentor's list to the mock server rather than using hardcoded data. 

**Here is a list of changes made:**
1. Clicking on "Find More Mentors" takes the user to the Explore page.
2. The mentor FullName and AvatarURL are now rendered from the mock server.

**Here is a code snippet where I would love to get a review on:**
Screenshot 2023-08-03 at 7 34 10 PM
This is my first time using the userProvider and putting in a Query. I have used the existing code in the mm-flutter-app repo for learning and as my reference. Here my assumption is that "data.response!.reversed" shows the list of recommended mentors in the order in which they have been recommended. I would love for the team to validate/correct this assumption for me. Thank you!

**Furthermore, I did not make changes to mm-mock-server/package-lock.json, package-lock.json, and package.json, so I am not sure why they're showing up here. I did use "npm install" on the root directly and "npm install package-lock.json" in the root directly, and I presume that there is something happening to do with that**

**Here is a list of future changes to come:**
1. Once the mentor bio and mentor skill fields show up in the mock server, tie those up to the widget.
4. Enable the recommended mentor's class to handle the scenarios where certain expected fields are not present, for instance avatar url, or bio, or skills. This requires changes in MentorCard.dart and RecommendedMentorsScroll.dart.
5. Is there a limit for the number of recommended mentor cards that can be shown in this top shelf? IF so, add that limit.


rupalg:
Update from Paris: the max number of recommended mentor cards that can be shown in this top shelf = 4.

ashlilili (dismissed):
LGTM, thanks for taking care of this!
For the extra generated files, you can either git rm <file> to remove them or git reset <file> to unstage them before committing but keep the local copies. I usually only git add the files I made changes to before committing to avoid this.

m-rosario:
I reverted the extra files. Thanks Rupal, this looks great 👍

m-rosario (approved):

PR #83, opened 2023-08-07
Add section "how can i help" for mentors

ashlilili:
Screenshot 2023-08-07 at 5 30 03 PM

Created "how can i help" section, including expertises/industries/availability/expectations subsections.
Using hardcoded chips, icon and text because we plan to retrieve these data from the backend in the future.

### DIFF
--- a/lib/widgets/atoms/profile_chip.dart
+++ b/lib/widgets/atoms/profile_chip.dart
@@ -6,24 +6,28 @@ class ProfileChip extends StatelessWidget {
   final String text;
   final IconData? icon;
   final bool? usePrimaryColor;
+  final bool? applyIconColor;
 
   const ProfileChip({
     super.key,
     required this.text,
     this.icon,
     this.usePrimaryColor,
+    this.applyIconColor,
   });
 
   @override
   Widget build(BuildContext context) {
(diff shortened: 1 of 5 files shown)


lib/widgets/atoms/profile_chip.dart
```diff
@@ -6,24 +6,28 @@ class ProfileChip extends StatelessWidget {
...
     this.icon,
     this.usePrimaryColor,
+    this.applyIconColor,
   });
 
   @override
   Widget build(BuildContext context) {
     final ThemeData theme = Theme.of(context);
+    Color iconColor = Color(theme.colorScheme.outline.value);
+    if (applyIconColor != null && applyIconColor!) {
```
m-rosario, line 23: nit: You can provide a default value for optional fields in the constructor like: `this.applyIconColor = true`, that way it will never be null and you don't have to do checks like these.

m-rosario (approved):
Looking great 😎

lib/widgets/atoms/profile_chip.dart
ashlilili: Great point! Thanks for the tip. I will add this fix in my next PR to save an approval here :)

PR #84, opened 2023-08-07
Mock server content and group updates

andersan:
Separated these changes from #73, which update the mock server to serve some information about users' expertises and industries.

Still need to add some logic to serve the country of residence, city of residence, and some other fields used in the Explore page.

EDIT: this is now ready for review, adds a bunch of fields to the mock server


andersan:
Should still add some more fields, like `CompanyType` and `CompanyStage` for mentees. Got caught up in some related language changes to the user model in the backend, so didn't get this in today.

LucasMagnum (approved):

PR #85, opened 2023-08-07
Inbox Chat state management with async queries and ChangeNotifier models

m-rosario:
This PR introduces my suggested approach for managing state using Provider and ChangeNotifier. The Inbox Chat was used as a proof of concept for this approach, meaning that only the MessagesProvider and part of the ChannelsProvider were affected by this change (UserProvider was not modified). The visible changes of this PR is that the Inbox Chat list and the 1-on-1 chat UI are updated in real time, without having to query the backend on every frame:

[chat-state-manage.webm](https://github.com/micromentor-team/mm-flutter-app/assets/25093428/98456250-efc2-4417-a328-45248bc1f5b5)

The following are the key changes in the structure:

- The classes that we call 'Providers' are now being used exclusively as an interface to GraphQL. This means that MessagesProvider is used only to execute GraphQL operations. It no longer manages Widgets or handles any state.
- 'Model' classes are introduced between the GraphQL operations and the Widgets that use the data. These models extend ChangeNotifier, and are used exclusively to manage state for specific UI component. Note that these models are different from database models (input and output objects to GraphQL queries), which are being represented by the Codegen autogenerated classes introduced a while back.
- Data is returned directly to the callers, which are then responsible for building Widgets to represent them. There are two scenarios presented in this PR: data managed by the ChangeNotifier model, and data that is simple enough to be returned directly to the Widget. In the latter case, a FutureBuilder is used to handle asynchronous operations.
- With this approach widgets don't make a query on every frame. Instead, a query is made whenever the widget is initialized and whenever there is an notification from the ChangeNotifier models, which in this case come from the Channel subscriptions.

Note that there is still a LOT of work to be done in order to optimize the requests and improve performance. This PR is just a sampler of the direction I propose.

Next steps once approved:
- Update all 'Provider' classes so that they all use asynchronous functions and don't handle any Widgets or state. None of them will extend ChangeNotifier anymore since they would only be used as clients to a subset of GraphQL operations.
- Reduce the overhead from the ChangeNotifier refresh() calls. Ideally, the subscription event should tell exactly what changed so that the refresh can be narrowed down as much as possible. Also, not all events should trigger a refresh.
- Find a way to implement cache efficiently and safely. There was a partial implementation before this PR, but it caused issues because the cache was not being updated for all mutations. Most mutations don't return the mutated object, therefore a second networkOnly query has to be made to get the updated objects, which defeats the purpose of the cache. Also, indiscriminate access to the cache results in outdated data, which is problematic as well.


ashlilili (approved):

lib/providers/messages_provider.dart
```diff
@@ -1,112 +1,89 @@
...
-      cache.writeQuery(
-        req,
-        broadcast: true,
-        data: response,
-      );
-    }
+  // Queries
+  Future<OperationResult<List<ChannelMessage>>> findChannelMessages({
+    required Input$ChannelMessageListFilter input,
+    bool fetchFromNetworkOnly = false,
```
LucasMagnum, line 23: Good job using the FetchPolicy, I got curious why not using only the FetchPolicy to control whether getting from network or not?

Would cacheAndNetwork policy give the same result? (https://github.com/zino-hofmann/graphql-flutter/blob/main/packages/graphql/lib/src/core/policies.dart#L27)

LucasMagnum (approved):
Thanks for this modification and your suggestion.

LucasMagnum:
> With this approach widgets don't make a query on every frame. Instead, a query is made whenever the widget is initialized and whenever there is an notification from the ChangeNotifier models, which in this case comes from the Channel subscriptions.

This is beautiful!

lib/providers/messages_provider.dart
m-rosario: cacheAndNetwork has to be used carefully because the returned data will always be stale. There are cases when this is ok, but chat is probably the one exception that always needs the latest data.

PR #87, opened 2023-08-08
Miscellaneous fixes in the home page

m-rosario:
Found a few bugs when testing with the real backend:

- The queryAllUsers call doesn't work without a 'filter' parameter. Added an inconsequential parameter just so that it won't crash and query all users.
- AvatarUrl in the Recommended Mentor tile was a required parameter, but in practice it can be null. Made it a nullable field.
- Added a default empty avatar image for header, recommended mentors, and upcoming sessions.

### DIFF
--- a/lib/widgets/atoms/mentor_card.dart
+++ b/lib/widgets/atoms/mentor_card.dart
@@ -3,7 +3,7 @@ import 'package:flutter_gen/gen_l10n/app_localizations.dart';
 import 'package:mm_flutter_app/constants/app_constants.dart';
 
 class MentorCard extends StatelessWidget {
-  final String avatarUrl;
+  final String? avatarUrl;
   final String mentorName;
   final String mentorBio;
   final List<String> mentorSkill;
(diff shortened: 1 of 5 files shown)


lib/widgets/atoms/mentor_card.dart
```diff
@@ -30,7 +30,9 @@ class MentorCard extends StatelessWidget {
             radius: Radii.avatarRadiusLarge,
             child: CircleAvatar(
               radius: Radii.avatarRadiusMedium,
-              backgroundImage: NetworkImage(avatarUrl.toString()),
+              backgroundImage: avatarUrl != null
+                  ? NetworkImage(avatarUrl!) as ImageProvider<Object>
```
jupblb, line 34: Do we need this casting here?

jupblb (approved):

lib/widgets/atoms/mentor_card.dart
m-rosario: Weirdly enough, yes, because of the ternary operator the type of the result cannot be determined... for some reason. 

`The argument type 'Object' can't be assigned to the parameter type 'ImageProvider<Object>?'`

Adding the cast to either result solves this.

PR #88, opened 2023-08-08
Hardcoded v1 of profile's about me section

rupalg:
Created "about me" section of the profile with hardcoded data. 

Ensured that formatting matched the layout on Ashley's implementation of the "how I can help section": https://github.com/micromentor-team/mm-flutter-app/pull/83

Currently, the country icon and the language icon doesn't match the Figma mocks. Is this necessary to match? There are no country icons on Flutter Icons widget.

Screenshot 2023-08-08 at 3 11 16 PM

### DIFF
--- a/lib/widgets/screens/profile/profile_about_me.dart
+++ b/lib/widgets/screens/profile/profile_about_me.dart
@@ -1,6 +1,5 @@
 import 'package:flutter/material.dart';
 import 'package:flutter_gen/gen_l10n/app_localizations.dart';
-
 import '../../../constants/app_constants.dart';
 import '../../atoms/profile_chip.dart';
 
@@ -79,6 +78,7 @@ class ProfileAboutMe extends StatelessWidget {
     final AppLocalizations l10n = AppLocalizations.of(context)!;
     final ThemeData theme = Theme.of(context);
 
+    //TODO: call the backend instead of using hardcoded data
     return Column(
       mainAxisAlignment: MainAxisAlignment.start,
       crossAxisAlignment: CrossAxisAlignment.start,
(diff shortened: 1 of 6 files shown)


lib/widgets/screens/profile/profile_about_me.dart
```diff
@@ -0,0 +1,156 @@
...
+        ]);
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final AppLocalizations l10n = AppLocalizations.of(context)!;
+    final ThemeData theme = Theme.of(context);
+
+    return Column(
+      mainAxisAlignment: MainAxisAlignment.start,
```
jupblb, line 83: You don't need to set this value to `MainAxisAlignment.start` as it's the default one. This applies to other parts of this PR where it's used.

lib/widgets/screens/profile/profile_about_me.dart
```diff
@@ -0,0 +1,156 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_gen/gen_l10n/app_localizations.dart';
+
+import '../../../constants/app_constants.dart';
+import '../../atoms/profile_chip.dart';
+
+class ProfileAboutMe extends StatelessWidget {
+  const ProfileAboutMe({Key? key}) : super(key: key);
```
jupblb, line 8: Nit: this may be simplified to `const ProfileAboutMe({super.key});`

jupblb (changes requested):
image
The chip sections is slightly misaligned when compared with the wireframe. Once https://github.com/micromentor-team/mm-flutter-app/pull/89 is merged you may want to use the `ProfileChipPadded` widget.

lib/widgets/screens/profile/profile_about_me.dart
```diff
@@ -0,0 +1,156 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_gen/gen_l10n/app_localizations.dart';
+
+import '../../../constants/app_constants.dart';
```
jupblb, line 4: Nit: it seems to me we are mostly using absolute imports throughout the app (`package:mm_flutter_app/...`).

lib/widgets/screens/profile/profile_about_me.dart
rupalg: Hi, okay- will remove this value wherever applied

lib/widgets/screens/profile/profile_about_me.dart
rupalg: okay

lib/widgets/screens/profile/profile_about_me.dart
rupalg: Good point. I am happy to keep the path absolute. I am seeing a combination of relative and absolute imports throughout the app though, so for future PRs, we can keep the path absolute.

PR #91, opened 2023-08-09
Update ChannelsProvider to use async queries

m-rosario:
- Update the remaining queries inside ChannelsProvider to return Future instead of Widget
- Updated the query for the channels last message to use instead Channel.latestMessage
- Some miscellaneous house keeping in related classes

Note that I tried to also implement an improved version of the query for unseen messages, but it seems that the corresponding backend query is not working as expected. This will be revised later.

### DIFF
--- a/lib/widgets/molecules/invitation_section.dart
+++ b/lib/widgets/molecules/invitation_section.dart
@@ -5,12 +5,14 @@ import 'package:mm_flutter_app/__generated/schema/schema.graphql.dart';
 import 'package:mm_flutter_app/constants/app_constants.dart';
 import 'package:mm_flutter_app/providers/channels_provider.dart';
 import 'package:mm_flutter_app/providers/user_provider.dart';
+import 'package:mm_flutter_app/utilities/utility.dart';
 import 'package:mm_flutter_app/widgets/atoms/invitation_tile.dart';
 import 'package:mm_flutter_app/widgets/atoms/section_tile.dart';
 import 'package:provider/provider.dart';
 
 import '../../__generated/schema/operations_channel.graphql.dart';
 import '../../__generated/schema/operations_user.graphql.dart';
+import '../../providers/base/operation_result.dart';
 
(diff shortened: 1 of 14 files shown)


lib/widgets/screens/channel_messages/channel_messages.dart
```diff
@@ -117,17 +136,16 @@ class _ChannelChatState extends State<ChannelChat> {
   final ScrollController listScrollController = ScrollController();
   final Duration _animationDuration = const Duration(milliseconds: 250);
   ChatModel? _chatModel;
+  MessagesProvider? _messagesProvider;
```
jupblb, line 139: You may mark it as `late final` and skip the `?`.

jupblb (dismissed):

lib/widgets/molecules/invitation_section.dart
```diff
@@ -21,43 +23,65 @@ class InvitationSection extends StatefulWidget {
...
+  void initState() {
+    super.initState();
+    _channelsProvider = Provider.of<ChannelsProvider>(context, listen: false);
+  }
+
+  @override
+  void didChangeDependencies() {
+    super.didChangeDependencies();
+    _invitationInbox = _channelsProvider?.getInboxInvitations();
+    l10n = AppLocalizations.of(context);
```
LucasMagnum, line 40: Is the context available here?

lib/widgets/screens/channel_messages/channel_messages.dart
m-rosario: You're super right! Replaced this and other declarations to use lazy initializer.

lib/widgets/molecules/invitation_section.dart
m-rosario: Yes, the BuildContext is an instance field of the State class. It is available right after initialization.

LucasMagnum (approved):

PR #94, opened 2023-08-10
Quick fix to change the padding of the chips in the "How can I help" section

rupalg:
Hello,
This is a quick fix to the padding of the "How can I help" section of the profile. Here's what it looks like on an Android emulator. 
Screenshot 2023-08-10 at 3 13 02 PM

### DIFF
--- a/lib/widgets/screens/profile/how_can_i_help_section.dart
+++ b/lib/widgets/screens/profile/how_can_i_help_section.dart
@@ -83,7 +83,8 @@ class HowCanIHelpSection extends StatelessWidget {
             children: [
               for (var chip in chips)
                 Padding(
-                  padding: const EdgeInsets.all(Insets.paddingExtraSmall),
+                  padding:
+                      const EdgeInsets.only(right: Insets.paddingExtraSmall),
                   child: chip,
                 )
             ],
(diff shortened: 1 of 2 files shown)


lib/widgets/screens/profile/how_can_i_help_section.dart
```diff
@@ -83,7 +83,8 @@ class HowCanIHelpSection extends StatelessWidget {
             children: [
               for (var chip in chips)
                 Padding(
-                  padding: const EdgeInsets.all(Insets.paddingExtraSmall),
+                  padding:
+                      const EdgeInsets.only(right: Insets.paddingExtraSmall),
```
m-rosario, line 87: This works for Left-to-right layouts, but remember that we also need to support Right-to-left. Generally speaking, never specify anything for 'left' and 'right' unless they both have the same value (because it won't matter if they get flipped).

For this use: `EdgeInsetsDirectional.only(start:  paddingExtraSmall),`

Flutter can determine the start or end of a widget of text depending the current local is Right-to-left or Left-to-right :-)

m-rosario (changes requested):

lib/widgets/screens/profile/how_can_i_help_section.dart
rupalg: That's super insightful. Let me make the change. Thanks Miguel

m-rosario: I'm so sorry, I meant to use `end: paddingExtraSmall` in my code snippet. In left-to-right, end is equal to right. That's how you know!

m-rosario (approved):
Thanks!

PR #98, opened 2023-08-13
Archiving/unarchiving of Inbox chats

m-rosario:
This PR implements functionality to archive/unarchive a channel either from the list of channels (with a swipe gesture) or from the channel messages screen (using three dots menu). For now all that archiving/unarchiving does is move the list in or out of the Archived list.

The mock server was updated to allow testing these implementations.


LucasMagnum (approved):
LGTM. 

Optionally I'd check with the UX how the empty state should look like (when there is no chats to show) + the icons to archive/unarchive.

PR #99, opened 2023-08-14
Firebase Hosting for Web deployment, doc+changes

andersan:
Set up a web deployment of the Flutter app, including docs and hard coded changes here. 

[This is deployed here](https://micromentor-d72ff.web.app/) and could be used to QA the app by team members without Android devices, to give us time to acquire those devices.

I thought about setting up a remote mock server, too, but it is probably better to only hook this up to a proper backend to increase visibility on the status of the backend+frontend integration.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,4 +59,10 @@ node_modules/
 /lib/firebase_options.dart
 /android/app/google-services.json
 /ios/Runner/GoogleService-Info.plist
+
+# Firebase Hosting config and built files
+.firebaserc
+firebase.json
+.firebase/
+
 *firebase_service_account.json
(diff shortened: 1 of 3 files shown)


andersan:
@LucasMagnum I'm interested in your thoughts about this. It's probably the simplest way to get something deployed and accessible to anyone who needs to test the app.

MicroMentor needs to have Flutter Web working. Maybe this will feel like a distraction from the primary goal of getting the Android app ready during the fellowship, but we'll need to get anything working in web anyway.

andersan:
will refactor this to use github actions like #108 to create a .env in firebase hosting

PR #100, opened 2023-08-14
Add support for subscriptions to mock server

m-rosario:
There is no out-of-the-box mock for subscriptions in the Apollo packages, so these were implemented using the WebSockets and an in-memory PubSub, as described in the [Apollo Server docs](https://www.apollographql.com/docs/apollo-server/data/subscriptions).

Tested with several operations on the Inbox chats.

### DIFF
--- a/mm-mock-server/src/mocks/mutations.ts
+++ b/mm-mock-server/src/mocks/mutations.ts
@@ -1,4 +1,5 @@
 import { MockServerState } from "./util/state";
+import { PUBSUB_OBJECT_CHANGED, PUBSUB_CHANNEL_CHANGED } from "./subscriptions";
 import * as generators from "./util/generators";
 
 export function mockMutations(serverState: MockServerState) {
@@ -21,6 +22,13 @@ export function mockMutations(serverState: MockServerState) {
         archiveChannelForMe: (_: any, args: { channelId: string }) => {
             let channel = serverState.channels.find((element) => element.id == args.channelId);
             channel.isArchivedForMe = true;
+            serverState.pubsub.publish(PUBSUB_OBJECT_CHANGED, { objectChanged: { objectId: args.channelId}});
+            serverState.pubsub.publish(PUBSUB_CHANNEL_CHANGED, { 
+                channelChanged: {
+                    channelId: args.channelId,
+                    eventType: 'channelUpdated',
+                }
+            });
             return channel.id;
         },
         createChannel: () => {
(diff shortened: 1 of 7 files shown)


mm-mock-server/src/mocks/mutations.ts
```diff
@@ -1,4 +1,5 @@
 import { MockServerState } from "./util/state";
+import { PUBSUB_OBJECT_CHANGED, PUBSUB_CHANNEL_CHANGED } from "./subscriptions";
```
andersan, line 2: This file is missing, maybe it's still in your local @m-rosario

mm-mock-server/src/server.ts
```diff
@@ -1,13 +1,21 @@
 import { ApolloServer } from "@apollo/server";
-import { startStandaloneServer } from "@apollo/server/standalone";
 import { addMocksToSchema } from "@graphql-tools/mock";
 import { makeExecutableSchema } from "@graphql-tools/schema";
+import { ApolloServerPluginDrainHttpServer } from '@apollo/server/plugin/drainHttpServer';
+import { expressMiddleware } from '@apollo/server/express4';
+import { WebSocketServer } from 'ws';
+import { useServer } from 'graphql-ws/lib/use/ws';
```
andersan, line 7: the above two libraries weren't added to the package.json (ws, graphql-ws). `cors` is also missing

andersan (changes requested):
The mock server isn't building with these changes.

I'm not sure how the output of these changes to the mock server will compare to MM's backend's output. I could manually do some tests and compare the output of the mock server to the output of the MM backend.

mm-mock-server/src/mocks/mutations.ts
m-rosario: Whoops, fixed.

mm-mock-server/src/server.ts
m-rosario: Whoops, fixed.

m-rosario:
Sorry about the missing files. It should be resolved now.

shadowbrush:
It builds for me now, and I tested it like this in the GraphQL Studio Sandbox (is that the correct name now?):

1. Create a channel with `createChannel`, and noted down the `id`
2. Add a `channelChanged` subscription using the above `id` for `channelId` (probably would work with any, right?)
3. Created a message with `createChannelMessage`
4. Saw the subscription message come in:

	    {
	      "data": {
	        "channelChanged": {
	          "channelId": "uLkrcBnlxQg4OabE8QPy6JJ1",
	          "eventType": "messageCreated",
	          "invitationId": "9750fa1b-42e7-495c-a30c-d6f6c429827f",
	          "messageId": "uLkrcBnlxQg4OabE8QPy6JJ1",
	          "participantId": "ba192417-192f-4c18-bcac-8bc8fd9c702f"
	        }
	      }
	    }

shadowbrush (approved):

PR #101, opened 2023-08-16
Fix provider avatar, home page findAllUsers bugs

andersan:
A couple changes which get the app into a more usable state when connected to the real backend.
 
1. ~~Hacky temp 'fix' on the provider avatar bug. The MMData backend doesn't provide URLs at this time for this field. As a more long term fix in the mmdata backend, we should provide a graphQL resolver for `avatarUrl` and rename the current field to `avatarPath` or something.~~ There is a [PR out](https://github.com/micromentor-team/mmdata/pull/286) in the backend which fixes this issue, not yet merged. This PR adds the fix for null avatar URLs used in `findAllUsers` to the `findMentorUsers` and `findMenteeUsers` queries.
2. Display only 10 Mentor users on the home page, rather than loading the entire users collection onto the app. The app is unusable when connected to the backend due to the amount of data returned.

The fix for 2 will eventually need to be refactored If I am a mentor, I'd probably want to see a list of mentees, not mentors. For that, we need to determine if we want to re-work the existing `RecommendedMentorsScroll` widget, maybe into a `RecommendedUsersScroll` that can contain either mentors, mentees, or a mix of both.

### DIFF
--- a/lib/providers/user_provider.dart
+++ b/lib/providers/user_provider.dart
@@ -148,7 +148,12 @@ class UserProvider extends BaseProvider with ChangeNotifier {
       response: queryResult.data != null
           ? Query$FindMenteeUsers.fromJson(
               queryResult.data!,
-            ).findUsers
+            ).findUsers.map((element) {
+              if (element.avatarUrl == "") {
+                return element.copyWith(avatarUrl: null);
+              }
+              return element;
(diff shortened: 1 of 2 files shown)


m-rosario (approved):
Thanks! 👍

PR #102, opened 2023-08-16
Improved event handling for channel messages

m-rosario:
Introduces channelChanged subscription event filtering so that models (a.k.a. ChangeNotifiers for UI components) only process events that are relevant to them. Since the type of event and channel/message id is now included in the subscription, event handling is now more specific and efficient:  we now refresh only the object that changed instead of refreshing all of the data for every event received.

Note that the frontend uses the models to implement optimistic updates, so that the UI is updated without querying the backend. An eventual 'refresh' with the backend data ensures that both backend and frontend are synced.

Includes several other fixes and improvements, but as indicated in the TODOs, there is still more work to do in regards to optimization.

### DIFF
--- a/lib/__generated/schema/operations_channel.graphql.dart
+++ b/lib/__generated/schema/operations_channel.graphql.dart
@@ -2336,6 +2336,13 @@ const documentNodeQueryFindChannelLatestMessage = DocumentNode(definitions: [
                 selectionSet: null,
               ),
               FieldNode(
+                name: NameNode(value: 'createdBy'),
+                alias: null,
+                arguments: [],
+                directives: [],
+                selectionSet: null,
+              ),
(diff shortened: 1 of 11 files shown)


andersan:
The error I flagged is also happening for me in the branch `ca/mmdata-backend-bugfixes` (with some local changes there, including the updated graphql schema), so I don't think it's caused by this PR. It may already be in `main`, or may be due to some change in the backend.

m-rosario:
Thanks for the review @andersan .

I haven't modified the MessageDetailsModal yet so I'm not sure what this error is about. One of my upcoming tasks is to get rid of this modal to implement it like the UX design:

image

I should be able to catch and resolve all these issues by then.

PR #105, opened 2023-08-18
Fix GroupMemberships on mock server

m-rosario:
The findUsers query with GroupMemberships was not working with the mock server. Some variables were misnamed and we also needed a resolver for the IGroupMembership type.


andersan (approved):
Good catches, Miguel. The `IGroupMembership` may have been a recent schema change, I can't remember.

PR #106, opened 2023-08-22
Made some updates based on the new design of login page including

ashlilili:
1. extracted social sign in buttons (will be reused for sign up page)
2. adjusted some details such as padding and text
Screenshot 2023-08-22 at 8 04 51 PM


LucasMagnum (approved):

m-rosario (changes requested):
Looks very slick, but did you use a Web browser to test this? When I run it in my Android emulator the vertical spacing seems excessive to the point I have to scroll to see all the buttons. 

Screenshot 2023-08-22 at 10 03 28 AM


I believe this is yet another instance of the difference in rendering between web and mobile versions. I used the Flutter Inspector to show the difference in padding side-by-side (Web left vs Android right):


Screenshot 2023-08-22 at 10 14 43 AM


Note all the extra padding in the Android version. I don't think we can get away from having to create multiple versions of the UI in order to accommodate different platforms and screen sizes, but for now we can at least focus on making it look good on phones.

ashlilili:
image

Thanks Miguel for testing it on Android. I removed the extra paddings, it now looks like above

m-rosario (approved):
Thanks!!

PR #108, opened 2023-08-24
Android app distribution

LucasMagnum:
* Add CD to build the app and release it internally


m-rosario (approved):
Looks great, thanks Lucas! It would be nice to eventually have a short description of this in the main README, since this is an Open Source project and other teams need to do their own little setup here as well.

andersan (approved):
Looks great, thank you Lucas!

LucasMagnum:
> Looks great, thanks Lucas! It would be nice to eventually have a short description of this in the main README, since this is an Open Source project and other teams need to do their own little setup here as well.

Totally agree Miguel, I am working on a Release.md doc that will describe this process a bit better when it`s done.

m-rosario:
@LucasMagnum I just noticed that the package name used is not consistent in this PR. In some instances 'com.micromentor.mm_flutter_app' is used, and in other instances it is 'org.micromentor.mm_flutter_app' (notice the org instead of com). I believe the correct package is the one that uses 'org' since that is the domain that MM owns.

This mismatch prevents the debug version from compiling. You probably didn't notice it in the release version because that one is probably using same one defined in the module.

I am letting you know instead instead of just making a merge myself because you probably have external dependencies in Play Console or elsewhere.

PR #109, opened 2023-08-25
Welcome page

ashlilili:
Created the UI of the welcome page and the change language page.
Added locale provider so the language could be changed dynamically.
Added a few new spanish and arabic translations for testing purpose.

[recording.webm](https://github.com/micromentor-team/mm-flutter-app/assets/100324646/133e0d77-1ef8-42e2-a0dc-62168775ff32)


lib/main.dart
```diff
@@ -76,13 +75,11 @@ class _StartScreenState extends State<StartScreen> {
...
           onLoading: () => const LoadingScreen(),
-          onError: () => SignInScreen(
-            nextRouteName: widget.nextRouteName,
-          ),
+          onError: () => const WelcomeScreen(),
           onReady: () {
             if (snapshot.data?.response == null) {
               // Empty result means that the user not signed in.
-              return SignInScreen(nextRouteName: widget.nextRouteName);
+              return const WelcomeScreen();
```
m-rosario, line 82: Here the nextRoutName param is lost. Lucas added this param so that users could open a screen directly from a notification. If the user is logged in, this _if_ statement is skipped and the WidgetsBinding callback below will make sure to redirect them to the corresponding page. However if the user is not logged in, this condition is executed to have them sign in and THEN redirect them to the nextRoute (which could be anything). 

By removing the nextRouteName param here the user will go through the sign in flow as usual and always end up in the home page. This is not super serious but I wanted to point this out since it renders the SignInScreen nextRoutName meaningless.

I would suggest something like this, but feel free to solve it however you wish:

```
if (snapshot.data?.response == null) {
    // Empty result means that the user not signed in.
    return widget.nextRouteName == Routes.home.name? 
      const WelcomeScreen() 
      : SignInScreen(nextRouteName: widget.nextRouteName;
}
```

This means: if the next route is home (opening the app normally) show the WelcomeScreen, otherwise the user just needs to sign-in to be redirected to a different page (coming from clicking a notification, or whatever).

m-rosario (changes requested):
This is amazing! Just left one comment!

lib/main.dart
ashlilili: Thank you for the explanation! I updated it to use the signin screen when nextRoutName is available.

m-rosario (approved):

PR #112, opened 2023-08-28
Update schema and reconcile breaking changes

m-rosario:
Updates the schema to the latest version. The following breaking changes were also addressed:

- ~~The User type no longer contains the **companies** field, so an additional query had to be made to build received invitation details.~~ Schema updated again so this is no longer needed.
- DateTimeISO type had to be specified in the codegen builder, and in the mock server.

EDIT: After the reverts and revisions the net effect of this PR is an updated schema, addition of DateTimeISO, and some minor formatting fixes. 🤷


lib/schema/schema.graphql
```diff
@@ -374,23 +371,26 @@ type User {
...
   """
   oneLiner: String
+  user: User!
   countryOfResidence: Country
+  gender: Gender
+  educationLevel: EducationLevel
+  channels: [Channel!]
+  channelInvitations: [ChannelInvitation!]
+  channelParticipants: [ChannelParticipant!]
+  companies: [Company!]
```
andersan, line 381: The Company array still seems to be supplied here - am I missing something?

andersan:
@m-rosario I'm not sure that the removal of `companies` from `User` was intentional. In the latest `main` of `mmdata`, I still see a `companiesField` on the GraphQL resolver for the User, so I would expect the functionality to be the same as previously.

The `DateTimeISO` change makes sense.

I re-generated the schema from the latest `main` in the backend repos, and the only difference from this PR is trivial:
```
input ModelEventInput {
  time: DateTimeISO! = "2023-08-28T22:40:36.416Z" << this date changed
```

lib/schema/schema.graphql
m-rosario: The query in question returns a User, not a UserExt

m-rosario:
@andersan If you plan on adding User.companies back then I can put this PR on hold. Let me know what you find!

andersan:
@m-rosario FYI #114 reverts some of these changes, i tested it locally in Chrome and the chat functionality seems to be unaffected (can send messages, archive/unarchive chats)

m-rosario:
@andersan Ok done! This PR should be ready for approval/merging. Thanks!

andersan (approved):
Thanks, Miguel!

PR #116, opened 2023-08-29
Add temporary logout button for testing purposes

m-rosario:
This adds a drawer with a logout button in the home page. The drawer appears after clicking the users avatar in the header.

Note that this is just a quick and dirty solution to allow us to switch accounts easily for testing purposes. Once UX provides us with the flow for signing out this will be reverted and replaced with that.

[Screen_recording_20230829_131848.webm](https://github.com/micromentor-team/mm-flutter-app/assets/25093428/307f737b-659f-4ebf-89a6-c2f2c51955d7)


andersan (approved):
Looks good, thanks for taking care of this detail Miguel!

PR #117, opened 2023-08-30
Fix bug that includes sent invites in list of received invites

m-rosario:
This will be revised later to use the new query "myChannelInvitations" instead of "myInbox", which allows querying sent and received separately.


andersan (approved):
Thanks for getting this in, Miguel. I should have verified on the backend what was provided.

After the demo, we could discuss modifying `pendingInvitations` to remove the ones sent by the user, unless that's needed elsewhere (e.g. in the home page).

PR #123, opened 2023-09-05
Sign up page and signup for email page UI

ashlilili:
Replaced the original sign up page with the first two screens of the new sign up page design. The web version looks like below: 

https://github.com/micromentor-team/mm-flutter-app/assets/100324646/9d3ca9d0-2600-4a78-a6dc-5cab82441315


m-rosario (approved):
Looks great on Android too!

Screenshot 2023-09-05 at 10 53 33 AM
Screenshot 2023-09-05 at 10 53 48 AM

In terms of code, while there is no right or wrong way to do it, I think that it would be easier to create a generic widget that has the scaffold and basic structure for all of the signup screens, and then pass in the parameters that make each screen unique (e.g. the the progress bar value, title, content, etc.)

Don't worry about changing this PR though! As you make the next couple of screens you will notice the parts that repeat and will see how they can be refactored!

Thank you for doing this! 🔥🔥

PR #126, opened 2023-09-07
Mentor Or Entrepreneur Screen with Radio Buttons

rupalg:
This PR creates a page in the sign up flow where the user can choose whether they're a mentor or a mentee using radio buttons. 

This PR does the following:
Screenshot 2023-09-06 at 5 48 36 PM
Screenshot 2023-09-06 at 5 48 46 PM

1) Routes the user using goRouter to this page in the sign up flow
2) Creates the custom UI for the Radio Button cards
3) Enables the behavior such that the selected card deselects the other
4) The selected card has a different background color

In the future, we can do the following:
1) Refactor this section such that a Radio Button Card class handles the different kinds of widgets, to make this widget more reusable
2) Navigate using the Navigator.


m-rosario (changes requested):
This looks amazing! However, remember we want to make the radio buttons a generic widget to avoid having to define all of this again in the other screen that uses them. 

Also, the focus of the task was in the component itself, not on the screen or routing. The screen structure (Scaffold, progress bar, buttons, etc) should be defined once in single generic widget, which I believe @ashlilili  should be taking care of. Otherwise, we would be having 20+ signup screens that contain mostly copy pasted code, and it would be a nightmare to maintain!

If you need an example of what I mean by 'generic widget', you can look into the ProfileQuickViewCard. It is a standalone component (not a screen) that is plugged into several parts of the app.

Let me know if you any help with this and thank you for coming up with this so quickly! 🔥

rupalg:
Hi Miguel, I have created a generic widget now. The widget is called "login_radio_button_cards.dart". This widget can handle both kinds of UIs that were in the Trello task. 

And the coolest part, phew, is that cards are grouped such that selecting one card deselects the others. This bit took forever. 

And of course, the selected card has a different background color. 

Screenshot 2023-09-07 at 4 56 24 PM


Screenshot 2023-09-07 at 4 56 48 PM

m-rosario (approved):
Awesome! The constructor for the cards might need more work later to avoid having to pass multiple lists that are related to a same object. 

I added a commit to remove this page from the nextRoute of the email page, since it is not the correct page that follows it. For testing purposes it's alright but wanted to avoid confusion since others are working on this same flow.

Thank you!

rupalg:
Yeah, I'd love to discuss with you how to iterate on the constructor down the line as well. Thanks!

PR #129, opened 2023-09-09
Update mock server resolvers and state to handle new inbox features

m-rosario:
Makes the latest app version functional with the mock server. The following were added:

- Added query for myChannelInvitations
- Server state starts with both sent and received invitations
- Server stat starts with 2 ongoing conversations, one with unseen messages, the other without
- Withdraw sent invitations

### DIFF
--- a/mm-mock-server/src/mocks/mutations.ts
+++ b/mm-mock-server/src/mocks/mutations.ts
@@ -109,6 +109,21 @@ export function mockMutations(serverState: MockServerState) {
             });
             return channelInvitation.id;
         },
+        deleteChannelInvitation: (_: any, args: { deletePhysically: boolean, channelInvitationId: string }) => {
+            const channelInvitation = serverState.channelInvitations.find((e) => e.id == args.channelInvitationId);
+            const currentTime : string = new Date().toISOString();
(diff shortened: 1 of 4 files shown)


mm-mock-server/src/mocks/mutations.ts
```diff
@@ -109,6 +109,21 @@ export function mockMutations(serverState: MockServerState) {
             });
             return channelInvitation.id;
         },
+        deleteChannelInvitation: (_: any, args: { deletePhysically: boolean, channelInvitationId: string }) => {
+            const channelInvitation = serverState.channelInvitations.find((e) => e.id == args.channelInvitationId);
+            const currentTime : string = new Date().toISOString();
+            channelInvitation.deletedAt = currentTime;
```
andersan, line 115: Is there already a way to test this in the UI?

andersan (approved):
Looks good, thank you Miguel!

mm-mock-server/src/mocks/mutations.ts
m-rosario: This is for the Withdraw functionality. And yes you can test it by opening a Sent invitation.

PR #134, opened 2023-09-13
Profile with backend integration, invitations header, and create invitations

m-rosario:
- Allow opening user profiles from the invitation detail card (sent or received)
- User data is fetched from the backend to populate the profile detail page sections. Note that several fields are still hardcoded because they are not defined in the backend. Once they are implemented these can be easily added to the profile.
- Added test profiles to the home page "recommended" section. Tapping on a card opens the profile.
- The user invitations are inspected to determine if there is a connection between the authenticated user and the user whose profile is being viewed. There are four possible states for the profile invitation header:

  1. **Nothing Shown**: Viewing my own profile, or another user's whose invitation has already been accepted/declined.
  2. **Cancel pending request**: Viewing profile of user that has not yet accepted/declined my invitation.
  3. **Accept/Decline**: Viewing profile of user that sent me an invitation to connect.
  4. **Send invitation**: Viewing profile of user I have not interacted with before.

The following video demonstrates the first three cases:

[profile-invites.webm](https://github.com/micromentor-team/mm-flutter-app/assets/25093428/026ac1a4-f819-4442-bd90-d59094b681a2)

Sending invitation:

[send-invite.webm](https://github.com/micromentor-team/mm-flutter-app/assets/25093428/e91e246c-46bc-4486-9e00-3b798d7ca615)


lib/widgets/molecules/recommended_mentors_scroll.dart
```diff
@@ -117,23 +121,47 @@ class _RecommendedSectionState extends State<RecommendedSection> {
...
   void didChangeDependencies() {
     super.didChangeDependencies();
-    _allUsers = _userProvider.findMentorUsers(
-      optionsInput: Input$FindObjectsOptions(limit: 10), // only return 10 users
-      filterInput: Input$UserListFilter(ids: null), // no filter/null filter
+    _recommendedMentors = _userProvider.findMentorUsers(
+      optionsInput: Input$FindObjectsOptions(limit: 20),
+      filterInput: Input$UserListFilter(
+        ids: [
+          // TODO - Test Users, remove when matching engine is ready
```
andersan, line 128: This is coming in #135

lib/widgets/screens/profile/about_my_business.dart
```diff
@@ -1,11 +1,16 @@
 import 'package:flutter/material.dart';
 import 'package:flutter_gen/gen_l10n/app_localizations.dart';
-import 'package:url_launcher/url_launcher.dart';
 import 'package:mm_flutter_app/constants/app_constants.dart';
 import 'package:mm_flutter_app/widgets/molecules/profile_chip_padded.dart';
+import 'package:url_launcher/url_launcher.dart';
 
 class AboutMyBusiness extends StatelessWidget {
-  const AboutMyBusiness({super.key});
+  final CompanyInput companyInput;
```
andersan, line 8: Is this type from the GraphQL types? If so, it should be a `Company` and not a `CompanyInput`. 

Generally, we only want to use the `Input` types when sending a mutation. This screen seems like it's meant to display the info about the company, not a form which we can use to update the company info.

lib/widgets/screens/profile/profile.dart
```diff
@@ -26,52 +119,198 @@ class ProfileScreenScroll extends StatefulWidget {
...
+                        g.industries.map((e) => e.translatedValue!).toList(),
+                    orElse: () => [],
+                  ),
+              mentoringPreferences: const [
+                "Weekly check-ins",
+                "Monthly check-ins",
+                "Informal chats",
+                "Formal meetings",
+                "Spot mentoring",
+              ], //TODO
```
andersan, line 259: Not yet implemented in the backend, not sure when this will be delivered

lib/widgets/screens/profile/profile.dart
```diff
@@ -26,52 +119,198 @@ class ProfileScreenScroll extends StatefulWidget {
...
+              expectations:
+                  'Sit amet justo donec enim diam vulputate ut pharetra sit amet aliquam id diam maecenas ultricies. Mi eget mauris pharetra et ultrices neque ornare aenean euismod elementum nisi quis eleifend.', //TODO
```
andersan, line 261: `mentorsGroupMembership.expectationsForMentees` should be used for this value, not sure if it's included in the schema yet or if it needs to wait on #135

lib/widgets/screens/profile/profile.dart
```diff
@@ -26,52 +119,198 @@ class ProfileScreenScroll extends StatefulWidget {
...
-            const ProfileAboutMe(),
-            const ExperienceAndEducation(),
-            const AboutMyBusiness(),
-            const HowCanIHelpSection(),
           ],
-        ),
+          const Divider(),
+          ExperienceAndEducation(
+            experience: [
+              ExperienceInput(
```
andersan, line 267: this is referred to as `BusinessExperience` in the backend, provided in #135

lib/widgets/screens/profile/profile.dart
```diff
@@ -26,52 +119,198 @@ class ProfileScreenScroll extends StatefulWidget {
...
+                position: "Founder",
+                companyName: "MC Consulting",
+                start: DateTime.utc(2016, 3),
+                end: DateTime.utc(2019, 5),
+                location: "Cleveland, Ohio, USA",
+                companyUrl: "https://mcconsulting.com",
+              ),
+            ],
+            education: [
+              EducationInput(
```
andersan, line 290: this is called `AcademicExperience`, also in #135

andersan: Fields are:
`institutionName`
`startDate`
`endDate` (optional)
`degreeType` (optional) -- i.e. "title" as listed here
`fieldOfStudy` (optional) --- i.e. "major" as listed here

lib/widgets/screens/profile/profile.dart
andersan: Fields:
`BusinessExperience.jobTitle` = `position` 
`BusinessExperience.businessName` = `companyName` 
`BusinessExperience.startDate`
`BusinessExperience.endDate` (optional)
`BusinessExperience.city`
`BusinessExperience.state`
`BusinessExperience.country`
^ maybe it makes more sense to combine the above into one `location` field, but I kept it consistent. Other location fields in the backend are all separating city/state/country. Once we know how these fields are defined in the onboarding/edit profile forms, we could consider combining these fields.

andersan (approved):
Looks great, thank you Miguel!

Some of these fields are provided in a draft PR i'm working on, so after that's merged, those areas can be updated.

Here is a [list of the fields](https://docs.google.com/spreadsheets/d/1N2AQhVO8mLLMHovYe-jsBfAw85Ei7iWcFnXOy5QWIsc/edit#gid=1374485682) (column F) that we'll be using. I believe this is more current than the PRD for the profile info, as I've been collaborating with Darin and Katy in this doc over the past week.

PR #135, opened 2023-09-14
Matching, profile updates to match backend progress

andersan:
Started these changes a couple of weeks ago while working on the backend's matching engine.

Goal:
- schema updates to match latest backend (done)
- updates to mock server for UserSearch api calls (partly done)
- updates to providers to allow for searches with filters (done)

TODO:
~~- add additional fields to the User provider (e.g. pronouns, cityOfOrigin, etc) that will be used on the detailed profile page~~
~~- remove changes to explore.dart screen, the scope of this PR is already too big~~

### DIFF
--- a/lib/schema/operations_user.graphql
+++ b/lib/schema/operations_user.graphql
@@ -43,13 +43,10 @@ query MyUserSearches {
         jobTitle
         countryOfResidence {
             translatedValue
-            textId
         }
         groupMemberships {
             groupIdent
             ... on MentorsGroupMembership {
-                expertisesTextIds
-                industriesTextIds
                 expertises {
                     translatedValue
                 }
@@ -59,8 +56,6 @@ query MyUserSearches {
                 endorsements
             }
             ... on MenteesGroupMembership {
-                soughtExpertisesTextIds
-                industryTextId
                 soughtExpertises {
                     translatedValue
                 }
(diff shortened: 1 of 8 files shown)


lib/schema/operations_user.graphql
```diff
@@ -146,6 +263,167 @@ query FindUserDetailedProfile($userId: String!) {
...
+  findUserById(id: $id) {
+    id
+    firstName
+    lastName
+    fullName
+    offersHelp
+    seeksHelp
+    websites {
+        value
+        label # label == "linkedin" for the linkedin button
```
andersan, line 276: Added some comments about the schema in this file

may need to refactor companies/businessExperiences, not sure about the requirements for that feature

andersan:
Column F of [this doc](https://docs.google.com/spreadsheets/d/1N2AQhVO8mLLMHovYe-jsBfAw85Ei7iWcFnXOy5QWIsc/edit#gid=1374485682) contains info about which fields/columns to use.

The mock server may still need to be updated, and some other schema changes may also come in the next few days. But the schema should be >90% correct for the purposes of the Explore and Profile pages (for the short term - long term, no idea how it'll change).

m-rosario (approved):
This is gold, thanks for taking the time to put this together!

PR #133, opened 2023-09-12
Part 1 of the multi-select chip

rupalg:
This widget allows the user to select a given number of options by tapping on the chips.

**Part 1 of the widget implements the following functionality**
- The two main parts of this widget are: text at the top of the list, and the list itself.
- The generic widget currently allows passing parameters for the title, and the list of chips (each chip has an optional icon and a required text).
Screenshot 2023-09-12 at 10 37 37 AM
Screenshot 2023-09-12 at 10 37 27 AM

**Part 2 of this widget would iterate on this chip to do the following:**
- the generic widget would allow passing the parameter for the maximum number of selections
- If the maximum number of chips is selected, tapping on additional chips has no effect. The user can always deselect a previous selection and then select a new one.

PR #136, opened 2023-09-14
Multi-select chips widget: retrieve the value for selected widgets

rupalg:
This PR implements the functionality such that "Tapping a chip selects it, and stores the value somewhere in the widget where it can be retrieved by the parent widget."

This is important for retrieving and sending the information from this form to the necessary db. You can test this functionality by commenting out these three lines in the build method:
    // var selectedChips = findSelectedChips(chipList);
    // var names = returnSelectedChipNames(selectedChips);
    // print(names);


rupalg:
The most recent commit ([e0e4e5f](https://github.com/micromentor-team/mm-flutter-app/pull/136/commits/e0e4e5fab7a17a0c96900562fa889045c45ddf25) enables the following:

- the generic widget would allow passing the parameter for the maximum number of selections
- If the maximum number of chips is selected, tapping on additional chips has no effect. The user can always deselect a previous selection and then select a new one.

With this functionality, it achieves all tasks needed for this widget as a follow up to https://github.com/micromentor-team/mm-flutter-app/pull/133

lib/widgets/molecules/multi_select_chips.dart
```diff
@@ -97,51 +153,49 @@ class CreateMultiSelectChips extends StatelessWidget {
   }
 }
 
-class FilterChipWidget extends StatefulWidget {
+class FilterChipWidget extends StatelessWidget {
+  final int id;
   final String chipName;
   final IconData? icon;
+  final bool selected;
+  final Function(int chipId) onSelectionChanged;
```
m-rosario, line 161: Each chip should also store the backend textId. The chipName is what we show to users, but remember that those change depending on the user language so we should not use that string value as an identifier.

For example, content from the backend will have a format similar to this:

```
{
    textId: 'accountingAndFinance',
    <... some other fields...>
    translatedValue: 'Accounting and Finance',
}
```
The **translatedText** ('Accounting and Finance') changes depending on the user's language, but the **textId** remains the same so that is what we use to identify these values.

To make long story short, we need a an additional field `textId` in the FilterChipWidget, and then the helper functions will return those textIds instead of the chipNames. Let me know if you need more clarification on this!

m-rosario (changes requested):
This is pretty good! Just left one suggestion for something that we will need when integrating with the backend.

lib/widgets/molecules/multi_select_chips.dart
rupalg: Hi Miguel, this is interesting! Thanks for pointing out. I have made the requested changes. PTAL

m-rosario (approved):
Thanks!!

PR #139, opened 2023-09-18
Connect Explore screen to a backend

jupblb:
This PR tries to make "Explore" screen show data retrieved from a backend instead of using hardcoded values. This is done by calling `FindRecommendedUsers` query (similar to `FindMentorUsers` and `FindMenteeUsers`) using `matchInput` generated from `ExploreCardFiltersModel`.

There are a few quirks about this PR that make it more of a "draft" right now:
1.  Since pagination is not implemented by any backend, support for this feature is skipped for now. Instead, each time a user hits the "See More" button, a new full list is retrieved with five more entries than before, replacing the previous list.

    Unfortunately, as I've tried to use `Future<OperationResult<T>>` pattern that I've observed in other widgets, it fails to work when its value is replaced. I don't know exactly why is that and could use some help figuring it out. @m-rosario

2.  There is some trouble with mapping all filter values to `Input$UserInput`. How to map country name to `countryOfOriginTextId`? There is the same issue with `spokenLanguagesTextId`. @andersan

3.  How to decide if a returned user is actually recommended? Should it be `false` by default for now, until another API is developed?

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -209,7 +209,7 @@ void main() async {
                   return ExploreCardFiltersModel(
                     countries: contentProvider.countryIds,
                     languages: contentProvider.languageIds,
-                    skills: contentProvider.expertiseIds,
+                    expertises: contentProvider.expertiseIds,
                     industries: contentProvider.industryIds,
                   );
                 },
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -177,6 +177,14 @@ void main() async {
               Provider<ContentProvider>.value(
                 value: ContentProvider(client: client),
               ),
+              FutureProvider(
+                initialData: ContentProvider(client: client),
+                create: (context) async {
+                  final provider = ContentProvider(client: client);
+                  await provider.findAllOptionsByType();
+                  return provider;
+                },
+              ),
               Provider<ChannelsProvider>.value(
                 value: ChannelsProvider(client: client),
               ),
@@ -192,8 +200,19 @@ void main() async {
               ChangeNotifierProvider(
                 create: (context) => ScaffoldModel(context: context),
               ),
-              ChangeNotifierProvider(
-                create: (context) => ExploreCardFiltersModel(),
+              FutureProvider(
+                initialData: ExploreCardFiltersModel.empty(),
+                create: (context) async {
+                  final contentProvider = ContentProvider(client: client);
+                  await contentProvider.findAllOptionsByType();
+
+                  return ExploreCardFiltersModel(
+                    countries: contentProvider.countryIds,
+                    languages: contentProvider.languageIds,
+                    skills: contentProvider.expertiseIds,
+                    industries: contentProvider.industryIds,
+                  );
+                },
               ),
               ChangeNotifierProvider(
                 create: (context) => LocaleModel(),
--- a/lib/providers/models/explore_card_filters_model.dart
+++ b/lib/providers/models/explore_card_filters_model.dart
@@ -45,7 +45,9 @@ class ExploreCardFiltersModel extends ChangeNotifier {
   bool get languageFilterSelected => _selectedLanguages.isNotEmpty;
   bool get expertiseFilterSelected => _selectedExpertises.isNotEmpty;
   bool get userFiltersSelected =>
-      expertiseFilterSelected || languageFilterSelected || countryFilterSelected;
+      expertiseFilterSelected ||
+      languageFilterSelected ||
+      countryFilterSelected;
 
   void setFilters({
     Set<String>? selectedCountries,
--- a/lib/providers/models/explore_card_filters_model.dart
+++ b/lib/providers/models/explore_card_filters_model.dart
@@ -81,14 +81,14 @@ class ExploreCardFiltersModel extends ChangeNotifier {
       languagesTextIds:
           _selectedLanguages.isEmpty ? null : _selectedLanguages.toList(),
       offersHelp:
-          (_selectedUserTypes.contains("Mentor") || _selectedUserTypes.isEmpty)
+          _selectedUserTypes.contains("Mentor")
               ? Enum$UserSearchFieldPreference.isTrue
-              : Enum$UserSearchFieldPreference.isFalse,
+              : Enum$UserSearchFieldPreference.any,
       searchText: selectedKeyword,
       seeksHelp:
           (_selectedUserTypes.contains("Mentee") || _selectedUserTypes.isEmpty)
               ? Enum$UserSearchFieldPreference.isTrue
-              : Enum$UserSearchFieldPreference.isFalse,
+              : Enum$UserSearchFieldPreference.any,
     );
   }
 }
--- a/lib/providers/models/explore_card_filters_model.dart
+++ b/lib/providers/models/explore_card_filters_model.dart
@@ -2,12 +2,25 @@ import 'package:flutter/material.dart';
 import 'package:mm_flutter_app/__generated/schema/schema.graphql.dart';
 
 class ExploreCardFiltersModel extends ChangeNotifier {
-  static const List<String> countries = ['USA'];
-  static const List<String> languages = ['English', 'Urdu', 'Hindi'];
-  static const List<String> skills = ['Marketing', 'Operations', 'StartingUp'];
+  final List<String> countries;
+  final List<String> languages;
+  final List<String> skills;
 
-  static const List<String> industries = ['Programming', 'Farming', 'Plumbing'];
-  static const List<String> userTypes = ['Mentor', 'Mentee'];
+  final List<String> industries;
+  final List<String> userTypes =  ['Mentor', 'Mentee'];
+
+  ExploreCardFiltersModel.empty()
+      : countries = [],
+        languages = [],
+        skills = [],
+        industries = [];
+
+  ExploreCardFiltersModel({
+    required this.countries,
+    required this.languages,
+    required this.skills,
+    required this.industries,
+  });
 
   Set<String> _selectedCountries = {};
   Set<String> _selectedLanguages = {};
--- a/lib/providers/models/explore_card_filters_model.dart
+++ b/lib/providers/models/explore_card_filters_model.dart
@@ -1,4 +1,5 @@
 import 'package:flutter/material.dart';
+import 'package:mm_flutter_app/__generated/schema/schema.graphql.dart';
 
 class ExploreCardFiltersModel extends ChangeNotifier {
   static const List<String> countries = ['USA'];
@@ -22,7 +23,10 @@ class ExploreCardFiltersModel extends ChangeNotifier {
 
   String? get selectedIndustry => _selectedIndustry;
   Set<String> get selectedUserTypes => _selectedUserTypes;
-  String? get selectedKeyword => _selectedKeyword;
+  String? get selectedKeyword =>
+      (_selectedKeyword == null || (_selectedKeyword?.isEmpty ?? false))
+          ? null
+          : _selectedKeyword!;
 
   bool get countryFilterSelected => _selectedCountries.isNotEmpty;
   bool get languageFilterSelected => _selectedLanguages.isNotEmpty;
@@ -50,5 +54,26 @@ class ExploreCardFiltersModel extends ChangeNotifier {
     _selectedUserTypes = selectedUserTypes ?? {};
     _selectedKeyword = selectedKeyword;
     notifyListeners();
+  }
+
+  Input$UserSearchInput toUserSearchInput(int maxResultsCount) {
+    return Input$UserSearchInput(
+      countryTextIds:
+          _selectedCountries.isEmpty ? null : _selectedCountries.toList(),
+      expertisesTextIds:
+          _selectedSkills.isEmpty ? null : _selectedSkills.toList(),
+      maxResultCount: maxResultsCount,
+      languagesTextIds:
+          _selectedLanguages.isEmpty ? null : _selectedLanguages.toList(),
+      offersHelp:
+          (_selectedUserTypes.contains("Mentor") || _selectedUserTypes.isEmpty)
+              ? Enum$UserSearchFieldPreference.isTrue
+              : Enum$UserSearchFieldPreference.isFalse,
+      searchText: selectedKeyword,
+      seeksHelp:
+          (_selectedUserTypes.contains("Mentee") || _selectedUserTypes.isEmpty)
+              ? Enum$UserSearchFieldPreference.isTrue
+              : Enum$UserSearchFieldPreference.isFalse,
+    );
   }
 }
--- a/lib/providers/models/explore_card_filters_model.dart
+++ b/lib/providers/models/explore_card_filters_model.dart
@@ -4,27 +4,27 @@ import 'package:mm_flutter_app/__generated/schema/schema.graphql.dart';
 class ExploreCardFiltersModel extends ChangeNotifier {
   final List<String> countries;
   final List<String> languages;
-  final List<String> skills;
+  final List<String> expertises;
 
   final List<String> industries;
-  final List<String> userTypes =  ['Mentor', 'Mentee'];
+  final List<String> userTypes = ['Mentor', 'Mentee'];
 
   ExploreCardFiltersModel.empty()
       : countries = [],
         languages = [],
-        skills = [],
+        expertises = [],
         industries = [];
 
   ExploreCardFiltersModel({
     required this.countries,
     required this.languages,
-    required this.skills,
+    required this.expertises,
     required this.industries,
   });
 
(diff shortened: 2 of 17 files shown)


andersan:
Thanks for getting these changes going on the Explore page! Some responses to the blockers:
1. Pagination was added a few months ago. I think it is supported by the `FindUsers` query in the Users provider. Here's a screenshot of an example query in postman:
Screenshot 2023-09-18 at 1 12 53 PM

The [recommended mentors molecule](https://github.com/micromentor-team/mm-flutter-app/blob/a23852fafadf98c1ce14b9d1142e98da71a28ae3/lib/widgets/molecules/recommended_mentors_scroll.dart) from the dashboard/home screen uses `limit` but not `skip`, in line 125.

2. @jupblb `FindUsers` is not intended to be used by the explore page in the long term, it was something we used when the explore page was first built. 

Instead, use `CreateUserSearch` and `FindUserSearchById`, which are added to the user provider in #135. 

Here's an example mutation with all of the filters included:
Screenshot 2023-09-18 at 1 11 16 PM

For paginating the users past the first 10, we might need to add a `getTopFoundUsers` query to the user provider.

3. I think this is resolved by using the `UserSearch` instead of `findUsers`, let me know if there's still any questions here.


FYI the remote server won't return any results in the `UserSearch` at the moment. We need to re-synchronize the users from our staging environment, and I was waiting to do that until I had some updates to Expertises and Industries completed.

lib/widgets/screens/explore/explore.dart
```diff
@@ -62,35 +115,76 @@ class _ExploreCardScrollState extends State<ExploreCardScroll> {
...
+    return FutureBuilder(
+      future: _recommendedUsers,
+      builder: (context, snapshot) => AppUtility.widgetForAsyncSnapshot(
+        snapshot: snapshot,
+        onReady: () => Column(
+          children: [
+            ..._createCards(snapshot.data?.response ?? []),
+            TextButton(
+                onPressed: () {
+                  setState(_loadMoreRecommendations);
```
m-rosario, line 127: setState fails to execute here because it cannot contain an async function. _loadMoreRecommendations() gets executed, but the widget does not rebuild so you cannot see the new results. You do not need to make changes here if you address my other comment.

lib/widgets/screens/explore/explore.dart
```diff
@@ -1,57 +1,110 @@
...
-      } else {
-        cardInfo.add(createRegularMentorExample());
-      }
-    }
+  static const int profileIncrement = 5;
+
+  int _profiles = 0;
+  Future<OperationResult<List<RecommendedUser>>>? _recommendedUsers;
+
+  Future<void> _loadMoreRecommendations() async {
```
m-rosario, line 37: You can make this `void _loadMoreRecommendations() { ... }` since you are not using `await` anywhere. Making this an async function breaks the setState pointed out in my other comment.

m-rosario (changes requested):
Looking good, left some comments that may help with the issue you had reloading results.

lib/widgets/screens/explore/explore.dart
jupblb: Changing the signature as requested doesn't seem to solve the problem. I'm not sure that `FutureBuilder` can be triggered for assigning another value to the digested `Future`.

jupblb:
This PR has been adapted to use the `CreateUserSearch` mutation. Issues #1 and #3 from the original PR message are no longer a problem. It's ready for the final review round.

One question left: how to map country/language to their respective IDs (`countryTextId`, `languagesTextId`)? @andersan

andersan:
@jupblb sounds great, will review this later today.

For the text IDs and the translated name of the country, those are accessible on the ContentProvider. Here's an example of how that might work:
```dart
    import 'package:mm_flutter_app/providers/content_provider.dart';

    late final ContentProvider _contentProvider;

    _contentProvider = Provider.of<ContentProvider>(
      context,
      listen: false,
    );

    // The content provider doesn't seem to have loaded the options anywhere. run this to load them
    await _contentProvider.findAllOptionsByType();

    print(_contentProvider.countryOptions?[0].textId);
    print(_contentProvider.countryOptions?[0].translatedValue);
```

I hope that code works, I haven't tested locally (just wrote it into the IDE for basic syntax checking...). The query is working on the remote backend to fetch countries, expertises and industries.

lib/providers/models/explore_card_filters_model.dart
```diff
@@ -22,7 +23,10 @@ class ExploreCardFiltersModel extends ChangeNotifier {
 
   String? get selectedIndustry => _selectedIndustry;
   Set<String> get selectedUserTypes => _selectedUserTypes;
```
andersan, line 25: The backend doesn't actually list the user types, but right now we just have two: "Mentor" and "Mentee". There isn't any plan in the scope to add more user types that can be searched for. (We do have `User.roles`, but that's different than the type of user. Mentor/mentee are not in that list of roles.)

Are these currently hardcoded somewhere here? I think having them in the front end is fine for now.

jupblb:
@andersan I've just updated this PR to use the `ContentProvider`. Please re-review if done already.

lib/providers/models/explore_card_filters_model.dart
jupblb: Yeah, this is the only value that remained hardcoded: https://github.com/micromentor-team/mm-flutter-app/pull/139/files#diff-1c59c7884f660b111a9fb79c05076a83e4ad06417087d8bea5b2782e30af5f3aR10

m-rosario (commented):
This is looking great so far! However I see that the mock server always returns the entire set of users for the search. Since we already have the search parameters in **args** we could filter the results to only return the users that satisfy the parameters. It doesn't have to be anything too fancy, I just figured that since we took the mock resolver this far we might as well push it a bit more to be like the real implementation.

Thank you for doing this, I will let Chris complete his review since he already has more context than I do.

lib/providers/models/explore_card_filters_model.dart
```diff
@@ -2,12 +2,25 @@ import 'package:flutter/material.dart';
 import 'package:mm_flutter_app/__generated/schema/schema.graphql.dart';
 
 class ExploreCardFiltersModel extends ChangeNotifier {
-  static const List<String> countries = ['USA'];
-  static const List<String> languages = ['English', 'Urdu', 'Hindi'];
-  static const List<String> skills = ['Marketing', 'Operations', 'StartingUp'];
+  final List<String> countries;
+  final List<String> languages;
+  final List<String> skills;
```
andersan, line 7: This should probably be renamed to `expertises` throughout the app, for consistency. Not sure where else `skills` is used or whether that's in scope for this PR.

lib/widgets/screens/explore/recommended_mentors_filters.dart
```diff
@@ -106,9 +115,15 @@ class _RecommendedMentorsFilters extends State<RecommendedMentorsFilters>
 }
 
 class _Expertise extends StatefulWidget {
-  final Set<String> skills;
+  final List<String> skills;
```
andersan, line 118: It looks like the parameter "skill" is what the backend and graphql models refer to as "textId". I'd consider renaming this for clarity, e.g. to expertiseTextId. (it's likely I've just become used to that naming convention, though, since it's what we use in the backend.)

lib/widgets/screens/explore/recommended_mentors_filters.dart
```diff
@@ -140,16 +155,16 @@ class _ExpertiseState extends State<_Expertise> {
               ? theme.colorScheme.onInverseSurface
               : theme.colorScheme.surface),
         ),
-        child: Text(l10n.exploreSearchFilterSkills(skill),
+        child: Text(widget.skillTranslate(skill) ?? skill,
             style: TextStyle(color: theme.colorScheme.primary)),
```
andersan, line 159: This doesn't seem to be working. Not sure what the cause is, but I can see that the UI is displaying the `textId` rather than the `translatedValue`:

Screenshot 2023-09-28 at 12 47 04 AM
Screenshot 2023-09-28 at 12 47 09 AM
Screenshot 2023-09-28 at 12 47 18 AM

lib/main.dart
```diff
@@ -170,6 +170,14 @@ void main() async {
               Provider<ContentProvider>.value(
                 value: ContentProvider(client: client),
               ),
+              FutureProvider(
+                initialData: ContentProvider(client: client),
+                create: (context) async {
+                  final provider = ContentProvider(client: client);
+                  await provider.findAllOptionsByType();
```
andersan, line 177: Looks good. I had to re-create the `Option`s, in particular `Language`s, in the remote server in order for this to work. The remote server was throwing an error because it had created these objects and stored them in MongoDB prior to the addition of `isUiLanguage`.

lib/widgets/screens/explore/recommended_mentors_filters.dart
andersan: I'm assuming this was not possible to test due to the issue with the remote server `Language`s mentioned in another comment

andersan (changes requested):
I'm happy to see the filters and the Content provider coming into this page!

Added some comments, including some semantics.

Here's a [new PR](https://github.com/micromentor-team/mm-flutter-app/pull/155) I added with a query to get paginated search results. 

IMO the use of the name "skill" rather than "expertise" is unfortunate for Micromentor, as we use the word "expertise" to refer to this concept. It appears to be in several other places in the app. Feel free to leave it however you see fit for now, and maybe in the future we can do an app-wide refactor.

lib/providers/models/explore_card_filters_model.dart
jupblb: Done. I've replaced it in some other places throughout this PR.

lib/widgets/screens/explore/recommended_mentors_filters.dart
jupblb: Good point. Renamed.

lib/widgets/screens/explore/recommended_mentors_filters.dart
jupblb: Ouch. Very true. I was able to test that but for some reason it worked back then. At the state of reviewed PR when an item is selected it'd be expanded into full, translated string. But for some reason this doesn't work for autoselect items.

jupblb: Fixed by adding: `displayStringForOption: optionsTranslations` in `Autocomplete`.

jupblb:
Thanks for the review @andersan. I've addressed your comments. The pagination feature is not implemented in this PR but for the sake of progress I suggest doing this in another one.

lib/providers/models/explore_card_filters_model.dart
```diff
@@ -51,4 +70,25 @@ class ExploreCardFiltersModel extends ChangeNotifier {
...
+      languagesTextIds:
+          _selectedLanguages.isEmpty ? null : _selectedLanguages.toList(),
+      offersHelp:
+          (_selectedUserTypes.contains("Mentor") || _selectedUserTypes.isEmpty)
+              ? Enum$UserSearchFieldPreference.isTrue
+              : Enum$UserSearchFieldPreference.isFalse,
+      searchText: selectedKeyword,
+      seeksHelp:
+          (_selectedUserTypes.contains("Mentee") || _selectedUserTypes.isEmpty)
+              ? Enum$UserSearchFieldPreference.isTrue
```
andersan, line 90: When I have nothing selected for the "User types" dropdown, it is setting both mentor and mentee preferences to `isTrue`. This means that it will only return users who are both mentors and mentees, which is a small percentage of the users. For example, only 6 users who speak English fulfill this condition in my local DB, which has a couple thousand users.

The scope here was probably not super specific on the behavior. 

One option for default values here could be:
- set `offersHelp` to `Enum$UserSearchFieldPreference.isTrue` by default if the user is a mentee (i.e. if the logged in user has `seeksHelp: true`). Also set `seeksHelp` to `Enum$UserSearchFieldPreference.any` in this case.
- if the user is not a mentee, set `seeksHelp` to `Enum$UserSearchFieldPreference.isTrue` and set `offersHelp` to `Enum$UserSearchFieldPreference.any` by default.

Let me know if that makes sense, thank you!

lib/widgets/screens/explore/explore.dart
```diff
@@ -1,57 +1,92 @@
...
-      if (cardInfo.isEmpty) {
-        cardInfo.add(createRecommendedMentorExample());
-      } else if (cardInfo.length == 1) {
-        cardInfo.add(createRecommendedEntrepreneurExample());
-      } else {
-        cardInfo.add(createRegularMentorExample());
-      }
-    }
+    _profiles += profileIncrement;
+    _recommendedUsers = widget.userProvider.createUserSearch(
```
andersan, line 38: This is a call that immediately returns a response, but the results aren't yet generated by the backend when it returns. A successful response to this mutation can be regarded as a sign that the backend is performing a search. `CreateUserSearch` will only return an error if the input to the search is somehow malformed or if the user has insufficient permissions.

We need to poll for `findUserSearchById` and check the `UserSearch.runInfos.finishedAt` to know if the search is complete. On that API call, we can use the `UserSearch`'s `topFoundUsers` to get the first 10 users.

After we know that a search is complete, for the paginated search results, the `findUserSearchResults` query can be used.

andersan: Just now I checked and verified the searches are actually finding results on the remote server, so that shouldn't be an issue.

andersan (changes requested):
Thanks for the quick updates, Michal. It looks like you addressed all the issues I raised in the last review.

Paginating the results in a future PR sounds good.

A few comments:
1. It seems that the autocomplete on the dropdowns is searching for the `textId` rather than the display value of the language. IMO that’s a bug. Especially for when we have translated language names. If a user is searching in Arabic, we’ll need to compare their Arabic script input to the Arabic language names.
2. Everything looks good with the mock server, but the remote backend isn’t working for me. It isn’t showing results when I first load the screen and also isn’t showing results after setting a filter. This is because the backend returns a successful call when `createUserSearch` is run, but it doesn’t return the results in that same call. I added more details in a comment on that line of the code. I could have caught this in the last review but missed it, sorry about that.
3. Out of scope for this PR, but noting here: when I view a profile that I already have a conversation with, their profile should have both a back button to go back to the explore page, and a button to continue the conversation with that user. When I ran the app, first profile shown by the mock server was a user who I already had a chat with, so that might be the easiest way to reproduce this.

lib/providers/models/explore_card_filters_model.dart
jupblb: I've switched this to:
```
          _selectedUserTypes.contains("Mentor")
              : Enum$UserSearchFieldPreference.any,
```
Setting the filter to counterpart of logged in user type still needs to be done.

lib/widgets/screens/explore/explore.dart
jupblb: I've not tested this with the real backend. This issue looks fair. I've for now added another `FutureBuilder` that fetches results. There is no polling yet, though.

jupblb:
Thank you for the review. I've yet to come up with solutions to some of the issues. For now I've fixed the autocomplete issue and added another query for the `findUserSearchById` to fetch results. Here are current TODOs:

1. Poll `findUserSearchById` to see if results are already available.
2. Support result pagination (may be done outside of this PR).
3. Set user type filter to the counterpart of the currently logged in user (e.g. if a mentee is signed in they will want to explore mentors by default; may be done outside of this PR).

jupblb:
Polling for users should now work.

andersan (approved):
Looks good, thank you Michal! The explore page is working for me with both a local and remote backend.

Some things to note, but which could be fixed later in separate PRs if easier:
1. Industry filters don't seem to be applied to the GraphQL query, and aren't displayed in this UI 
Screenshot 2023-10-02 at 12 01 30 PM

2.  Sometimes, the explore page is empty without a clear reason why. When first opening the screen, it usually populates with users for me but sometimes does not. It also occasionally doesn't display anything after a user search, even though the backend did find some users and there is no exception thrown by Flutter. Not sure of the cause of this issue or whether it happens for other users. I noticed this both running a local backend and local db, so minimal latency, and with the remote backend.
3. Clearing the "normal" filters doesn't clear the advanced filters. Not sure what the expected behavior here is, and it likely hasn't been defined in the requirements.
4. The "Expertises" appear to be displayed in the "Industry" advanced filters dropdown. "Volunteer Management" is an example of an Expertise, and "Furniture" is an example of an industry.
5. There is no UI displayed when there are no users found. If the latest `runInfo` of a `UserSearch` has a non-null `finishedAt` and has a `matchCount` of 0, we could display some UI that indicates that there were no users found by the search.

PR #141, opened 2023-09-20
Add detailed profile fields to mock server and app user profiles

m-rosario:
Add new User fields to the mock server and the detailed profile page. Only a few fields remain to be defined.


lib/constants/enums.dart
```diff
@@ -16,3 +16,16 @@ enum MessageDirection {
   received,
   unset,
 }
+
+enum GroupIdent {
```
andersan, line 20: There are more group idents in addition to these, but these are the "Core groups" for mentors/mentees. 

IMO this is fine, just wanted to add a note for context

lib/widgets/screens/profile/profile.dart
```diff
@@ -177,36 +186,39 @@ class _ProfileScreenScrollState extends State<ProfileScreenScroll> {
...
+            pronouns: userData.pronounsDisplay.isNotEmpty
+                ? userData.pronounsDisplay
+                : null,
             affiliations: const ["Verizon Digital Ready"], //TODO
             company: company.name,
-            companyRole: widget.userData.jobTitle,
-            education: widget.userData.educationLevel?.translatedValue,
-            linkedinUrl: "https://www.linkedin.com/in/williamhgates/", //TODO
+            companyRole: userData.jobTitle,
+            education: userData.educationLevel?.translatedValue,
```
andersan, line 199: Is this used at all? is it for titles (e.g. "Dr. First Last")?

lib/widgets/screens/profile/profile.dart
```diff
@@ -215,100 +227,74 @@ class _ProfileScreenScrollState extends State<ProfileScreenScroll> {
...
+                    [],
                 mission: company.description,
                 imageUrls: [
                   "https://st2.depositphotos.com/1326558/7163/i/600/depositphotos_71632883-stock-photo-mexican-tacos-with-meat-vegetables.jpg",
                   "https://st3.depositphotos.com/13349494/32631/i/600/depositphotos_326317470-stock-photo-cropped-view-man-adding-minced.jpg",
-                ],
-                motivation:
-                    'Sit amet justo donec enim diam vulputate ut pharetra sit amet aliquam id diam maecenas ultricies. Mi eget mauris pharetra et ultrices neque ornare aenean euismod elementum nisi quis eleifend.', //TODO
+                ], //TODO
+                motivation: //TODO
```
andersan, line 240: `motivationsForMentorship` is on the MenteesGroupMembership

andersan (changes requested):
Looks great, thank you Miguel! Left a few comments with minor changes

lib/widgets/screens/profile/profile.dart
```diff
@@ -177,36 +186,39 @@ class _ProfileScreenScrollState extends State<ProfileScreenScroll> {
...
             vacationMode: true, //TODO
           ),
           const Divider(),
           ProfileAboutMe(
-            cityOfResidence: widget.userData.cityOfResidence,
-            countryOfResidence:
-                widget.userData.countryOfResidence?.translatedValue,
-            cityFrom: null, //TODO
-            countryFrom: null, //TODO
+            cityOfResidence: userData.cityOfResidence,
```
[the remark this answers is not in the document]
m-rosario: Added!

lib/widgets/screens/profile/profile.dart
m-rosario: Added!

lib/widgets/screens/profile/profile.dart
m-rosario: @andersan 

We are displaying this field in the profile underneath the job title. I am not sure if we should use the latest business experience and academic experience for these fields, or if the user will be able to pick the job and school to show here.

It is still not super clear to me how the User.jobTitle differs from the BusinessExperience.jobTitle, and which one to use in the quick view card.

andersan: We may combine companies and `BusinessExperience` later on and just have one object type, but the requirements are a bit unclear. If we do combine those, we can remove `User.jobTitle`.

Hopefully after the onboarding flow is defined it'll be clearer how this data should be organized.

andersan (approved):
Thanks for making those changes!

PR #143, opened 2023-09-20
Profile Quick View card UI changes

rupalg:
This PR implements the changes in the UI of the profile quick view card as seen below
Screenshot 2023-09-20 at 2 41 02 PM
Screenshot 2023-09-20 at 2 41 09 PM

It is worthwhile to note that with the UI changes:
- we do not have any UI elements to show whether a profile widget is "recommended" for the user or not
-  we do not have the "venture stage" and "venture industry" sections as we did before for the entrepreneurs
- Before,  entrepreneurs did not have endorsements, but the redesign suggests that the entrepreneurs can have endorsements too

To reflect these changes, I have added new fields to and removed redundant fields from the Profile Quick View Card widget.


rupalg:
Note that the two fields needed to bring the fields into the present card are not in the backend yet.

These fields are:

endorsements for entrepreneurs
a field that notes whether the mentor is verizon digital ready or not
‌

Until those fields are added, no change is needed. @andersan

andersan:
For the endorsements, since I think they are P2 and we have some pending changes to endorsements in our current Django app, I'll wait to make those updates on the backend.

For the digital ready field, I think we should use the name of any Group that the user belongs to. If they are part of any group other than the mentees and mentors group, we can just show the first Group's name. I'll add a graphql field `User` to provide that logic, something like `groupNameToDisplay`.

lib/widgets/atoms/skill_chip.dart
```diff
@@ -9,7 +9,10 @@ class SkillChip extends StatelessWidget {
   const SkillChip({
     super.key,
     required this.skill,
-    this.icon,
+    this.icon = const Icon(
+      Icons.campaign_outlined,
+      color: Color(0xFF1D192B),
```
m-rosario, line 14: Avoid hardcoded colors. This will cause issues if, for example, we use dark theme, or just switch to a different color scheme in general.

lib/widgets/molecules/profile_quick_view_card.dart
```diff
@@ -238,96 +206,93 @@ class ProfileQuickViewCard extends StatelessWidget {
...
+            Expanded(
+              child: Column(
+                mainAxisAlignment: MainAxisAlignment.start,
+                crossAxisAlignment: CrossAxisAlignment.start,
+                children: [
+                  if (verizonDigitalReady == true)
+                    ProfileChip(
+                      text: l10n.verizonDigitalReady,
+                      verizonChip: true,
+                    ),
```
m-rosario, line 261: The chip in general displays group memberships other than mentor/mentee. Verizon Digital Ready is just used as an example in the mock. The backend for this still needs some work, but we should at least have this chip be customizable to display whatever String we get from the query

m-rosario (dismissed):
Thanks Rupal, left some comments. I ran this on Android and the padding/alignments need a bit more work. In particular notice the arrow in the top right and the text next to the avatar:

Screenshot 2023-09-20 at 6 40 45 PM

lib/widgets/molecules/profile_quick_view_card.dart
rupalg: okay

rupalg:
Thanks for the review. I will make the changes. 

What is the issue with the text next to the avatar @m-rosario ?

m-rosario:
> Thanks for the review. I will make the changes with the arrow's padding.
> 
> What is the issue with the text next to the avatar in the Android @m-rosario ?

Thanks! The icons have too much padding on the left so it doesn't look aligned with the text for the job title. But it's fine we can revise later let's not spend too much time here

m-rosario (approved):

PR #144, opened 2023-09-21
Dialogs for Block and Report user

m-rosario:
Adds the UI for the Block and Report dialogs. These can be accessed from a received invitation, chat UI, or in the detailed user profile.

[user-popup-menus.webm](https://github.com/micromentor-team/mm-flutter-app/assets/25093428/e3962a7f-4c3e-4adc-801e-bba9a6c0d35b)


lib/widgets/screens/inbox/channel_messages/channel_messages_screen.dart
```diff
@@ -66,16 +66,15 @@ class _ChannelMessagesScreenState extends State<ChannelMessagesScreen>
...
                 .firstWhere((item) => item.user.id != _authenticatedUser.id)
                 .user;
-            final String channelName = participant.fullName!;
-            final String? avatarUrl = participant.avatarUrl;
             buildPageRouteScaffold((scaffoldModel) {
               scaffoldModel.setParams(
                 appBar: AppBarFactory.createChannelMessagesAppBar(
                   context: context,
-                  channelName: channelName,
+                  userFullName: participant.fullName!,
```
jupblb, line 73: I think it's better to use `?? "Unknown user"` here. There have been cases in the past of using `!` with fields that were unsupported by the mock server.

jupblb (approved):
LGTM!

PR #145, opened 2023-09-21
Add capabilities and other minor ios config changes

andersan:
Added some changes made here while following some guides to set up an iOS deployment manually, which I see as a precursor to getting the CI/CD set up for iOS. (guides: https://docs.flutter.dev/deployment/ios, https://firebase.flutter.dev/docs/messaging/apple-integration#next-steps, https://firebase.google.com/docs/cloud-messaging/ios/client)


andersan:
The app is now working in my iPhone via TestFlight.

We need to assess whether we'd be able to get this deployed for other iPhone users for the testing session in 2.5 weeks. I am able to access the build via my Mercy Corps Apple Developer account, but for "external testers" without Apple Developer accounts within the Mercy Corps Apple account, we may need the app to go through a review from Apple.

m-rosario (approved):
Thanks!

PR #148, opened 2023-09-22
Entrepreneur flow pt 1

rupalg:
This is part of the sign up flow where the entrepreneur flow begins. 

For testing purposes, to trigger this flow, you can trigger "context.push(Routes.signupEntrepreneurOrMentor.path);" on any press event. 

Currently, it comes from the sign_up_community_guidelines.dart page.

Here's a screenrecording:

https://github.com/micromentor-team/mm-flutter-app/assets/63002494/71483321-66c9-4361-9a58-3f3920395202


rupalg:
Do note that the bottomBar where we have the "This will be shown on your profile" widget is part of the "sign_up_icon_footer.dart". I tried playing with it but couldn't figure out how to change the UI such that the text is centered. @ashlilili can you take that on? Thanks!

ashlilili:
> Do note that the bottomBar where we have the "This will be shown on your profile" widget is part of the "sign_up_icon_footer.dart". I tried playing with it but couldn't figure out how to change the UI such that the text is centered. @ashlilili can you take that on? Thanks!

Ah thanks for pointing this out, I also just noticed this! I will have a PR to fix this today :)

ashlilili (approved):
Thanks!

PR #151, opened 2023-09-27
Signup flow: gender, location, languages

m-rosario:
Adds pages to the signup flow, connecting to the user specific pages that were already built.

Some additional improvements in this PR:

- The TextFormFieldWidget was revised to use an outlined border like the one used in the UX design
- SignUpTemplate had some tweaks to prevent the scaffold from resizing each time the keyboard appeared on screen, which caused a lot of overflow errors. A GestureDetector was also added to make taps outside of interactible widgets refocus the app and dismiss the on-screen keyboard.
- AutocompleteWidget was revised to make it match the latest UX design.
- Initialized ContentProvider on StartScreen, which is the first screen launched by the app. Loading state is shown until the content provider has loaded all options.

Note that these pages still need more work. For example, the location page should have an autocomplete functionality to allow the user to select the city and country with ease. These are skipped for now until we have a basic working version of the signup flow.

[gender-location-lang.webm](https://github.com/micromentor-team/mm-flutter-app/assets/25093428/14565597-eaf1-4df9-bca7-bc285b12aa92)


rupalg (approved):
Thanks Miguel!

PR #153, opened 2023-09-27
Navigation of the mentor screens

rupalg:
> Creating this PR with the navigation across the different mentor screens
> TODO: Populate these screens with the necessary widgets


rupalg:
Added the necessary widgets within the mentor sign up flow:
https://github.com/micromentor-team/mm-flutter-app/assets/63002494/84e481cb-eab7-4b59-a99e-1ac77c592653

Next item: Insert the logic such that the user goes into a flow based on what role they have, i.e. mentor or entrepreneur.

lib/constants/routes.dart
```diff
@@ -121,6 +121,42 @@ class Routes {
...
+    path: '${signup.path}/signupMentorMoreInfo',
+  );
+  static Route signupMentorProfilePic = Route(
+    name: 'signupMentorProfilePic',
+    path: '${signup.path}/signupMentorProfilePic',
+  );
+  static Route signupMentorPronouns = Route(
+    name: 'signupMentorPronouns',
+    path: '${signup.path}/signupMentorPronouns',
+  );
```
m-rosario, line 139: These three pages are pretty much the same as the ones for entrepreneur. It is a waste to recreate them, instead we should reuse the existing ones.

Determining which page goes next and whether to display  "Find mentors" or "Find entrepreneurs" in the MoreInfo left button can be done by inspecting the current user to determine if they are a mentor or an entrepreneur.

I am aware that we still don't have the logic to pass data between these pages, but I am just letting you know that we should not spend any time creating duplicate pages like these.

m-rosario (changes requested):

m-rosario (approved):
Agreed to delete these later.

PR #154, opened 2023-09-28
Entrepreneur signup flow with content provider

m-rosario:
Fetches options from the backend using the content provider. This PR integrates the expertise picker, the pronoun picker, and the industry picker. This should cover the entire entrepreneur flow.

Note that some functions with hardcoded data were left in some classes because these are being used in the pages that Rupal is working on, so I chose not to delete them to avoid conflicts. These should be cleaned up as soon as we integrate the mentor's flow with the content provider.

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -59,7 +59,7 @@ class StartScreen extends StatefulWidget {
 
 class _StartScreenState extends State<StartScreen> {
   late Future<OperationResult<AuthenticatedUser?>> _authenticatedUser;
-  late Future<OperationResult<AllOptionsByType>> _optionsByType;
+  late Future<OperationResult> _content;
   late final InboxModel _inboxModel;
   late final ContentProvider _contentProvider;
 
@@ -73,7 +73,9 @@ class _StartScreenState extends State<StartScreen> {
   @override
   void didChangeDependencies() {
     super.didChangeDependencies();
-    _optionsByType = _contentProvider.findAllOptionsByType();
+    _content = _contentProvider.findAllOptionsByType().then(
+          (value) => _contentProvider.findPresetPronouns(),
+        );
     _authenticatedUser =
         Provider.of<UserProvider>(context).getAuthenticatedUser(
       logFailures: false, // Error is expected when user is not logged in.
@@ -87,7 +89,7 @@ class _StartScreenState extends State<StartScreen> {
   @override
   Widget build(BuildContext context) {
     return FutureBuilder(
-      future: _optionsByType,
+      future: _content,
       builder: (_, optionsSnapshot) => AppUtility.widgetForAsyncSnapshot(
         snapshot: optionsSnapshot,
         onReady: () => FutureBuilder(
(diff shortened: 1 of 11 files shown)


lib/widgets/screens/sign_up/sign_up_business_help_selection.dart
```diff
@@ -17,27 +20,51 @@ class SignupBusinessHelpSelectionScreen extends StatefulWidget {
...
 class _SignupBusinessHelpSelectionScreenState
     extends State<SignupBusinessHelpSelectionScreen> {
+  late final ContentProvider _contentProvider;
+  late final List<SelectChip> _expertiseChips;
+
+  @override
+  void initState() {
+    super.initState();
+    _contentProvider = Provider.of<ContentProvider>(context, listen: false);
+    _expertiseChips = _contentProvider.expertiseOptions!
```
andersan, line 30: There only seem to be a few expertises displayed here, is that the expected behavior?

Not sure if we were supposed to show all industries/expertises in the onboarding flow

lib/widgets/screens/sign_up/sign_up_mentor_pronouns.dart
```diff
@@ -50,3 +51,81 @@ class _SignupMentorPronounsScreenState
...
+          selectionOrder: _selectionorder1,
+          onChanged: (bool value) {
+            setState(() {
+              _isSelected1 = value;
+              if (value == true) {
+                _selectionorder1 = _numPronounsSelect + 1;
+                _numPronounsSelect = _numPronounsSelect + 1;
+              } else {
+                _selectionorder1 = _numPronounsSelect - 1;
+                _numPronounsSelect = _numPronounsSelect - 1;
```
andersan, line 87: This is interesting, I hadn't considered that we need to keep track of the order in which the user selected their pronouns.

The backend will store this as an array of `Pronoun` "textId"s, so if it is sent ['f','x'] or ['x','f'], it will keep the order sent by the Flutter app.

andersan (approved):
Looks great! I'm not 100% sure about the expertises/industries and how they compare to the PRD/wireframes. Other than that, this looks good to me.

Thank you, Miguel!

lib/main.dart
```diff
@@ -73,7 +73,9 @@ class _StartScreenState extends State<StartScreen> {
   @override
   void didChangeDependencies() {
     super.didChangeDependencies();
-    _optionsByType = _contentProvider.findAllOptionsByType();
+    _content = _contentProvider.findAllOptionsByType().then(
```
andersan, line 76: Nit: We might want to check for the content options via getters first, e.g. `presetPronounOptions`. If we only need the Pronoun options or if all of the options are already loaded, this isn't ideal. 

I think all the options are something like 100-200kb of uncompressed data (in json format) at the moment, so this is not a big deal.

lib/widgets/screens/sign_up/sign_up_business_help_selection.dart
m-rosario: The UI flow is hardcoded to show pages for the mentor, so you are not seeing the pages I worked with. I did not want to change that to prevent conflicts with Rupal's work. It is a bit messy right now but trust me we will clean this up asap

lib/main.dart
m-rosario: This query is triggered once per app start. I figured it would be cleaner to have query done once and then elsewhere we can assume the data is there, otherwise we need to put checks and queries everywhere else we plan to use these values. But yes maybe we can look into how we can optimize this further.

PR #155, opened 2023-09-28
Add paginated user search results to user provider

andersan:
Fixes an issue which may be blocking #139, the lack of paginated user search results.

`topUserIds` from the searches only included the top 10 users from a search. The search results exposed here will allow us to get more results. Right now, it allows us to fetch the entire list, so I recommend using some variables like this to limit the response to a reasonable number at first:

```json
{
    "userSearchId":"6514ab0174d1baabebddca7c",
    "options": {
        "limit": 10,
        "skip": 0
    }
}
```

### DIFF
--- a/lib/__generated/schema/operations_content.graphql.dart
+++ b/lib/__generated/schema/operations_content.graphql.dart
@@ -3,7 +3,6 @@
 // ignore_for_file: type=lint
 
 import 'package:gql/ast.dart';
-
 import 'schema.graphql.dart';
 
 class Query$FindAllOptionsByType {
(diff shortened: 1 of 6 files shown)


m-rosario (approved):
Thanks! 

I see that this schema update also includes operations for blocking and determining relationships. Can you confirm these functionalities are ready for use? Regarding the relationships attribute, is this to be used to determine if given users are connected, pending invitation, etc?

andersan:
@shadowbrush could you chime in here re: blocking and relationships? 

I'm not sure if we're still working on anything here, or if we've completed development & testing on blocking+reporting and are ready to integrate. Should we do some end-to-end testing on these flows via Postman to make sure it's ready?

andersan:
@m-rosario Holger said this is ready for integration. May not be bug-free, but it's almost there if not 100% complete.

PR #157, opened 2023-09-28
Home page improvements 

rupalg:
This PR:
- Revised header with welcome message and overflow menu (still not defined)
- Revised profile completion reminder banner

Here's how it looks: [look at the top section only]
Screenshot 2023-09-28 at 4 52 00 PM

One thing to confirm: The figma notes that the reminder banner color on Inverse Surface, and that color looks different in app than it does on Figma. That's all!


lib/widgets/atoms/profile_header.dart
```diff
@@ -18,52 +16,44 @@ class ProfileHeader extends StatelessWidget {
...
-                  ),
-                ),
-              InkWell(
-                onTap: () {
-                  //TODO(m-rosario): Remove this once the real logout flow is defined by UX
-                  Provider.of<ScaffoldModel>(
-                    context,
-                    listen: false,
-                  ).setTestHomeScaffold();
-                  Scaffold.of(context).openDrawer();
```
m-rosario, line 57: It would be great if you included this in the revision so that we have a way to logout during testing.

lib/widgets/atoms/profile_header.dart
```diff
@@ -18,52 +16,44 @@ class ProfileHeader extends StatelessWidget {
...
-            child: Padding(
-              padding: const EdgeInsets.all(Insets.paddingExtraSmall),
-              child: Text(
-                profileMessage,
-                textAlign: TextAlign.start,
-                softWrap: true,
-                style: theme.textTheme.titleLarge?.copyWith(
-                  color: theme.colorScheme.onSurface,
-                ),
-              ),
```
m-rosario, line 37: Without this Flexible widget there is risk of overflow. See this example:

Screenshot 2023-09-28 at 6 58 51 PM

m-rosario (changes requested):
Great job doing this so quickly! Left some comments.

lib/widgets/atoms/profile_header.dart
rupalg: Sounds good!

lib/widgets/atoms/profile_header.dart
```diff
@@ -18,52 +19,52 @@ class ProfileHeader extends StatelessWidget {
...
-                },
-                child: CircleAvatar(
-                  radius: Radii.avatarRadiusSmall,
-                  backgroundImage: avatarUrl != null
-                      ? NetworkImage(avatarUrl!) as ImageProvider<Object>
-                      : const AssetImage(Assets.blankAvatar),
-                ),
-              ),
-            ],
+          const Spacer(),
```
m-rosario, line 55: Wrapping the Text widget with a Flexible solves the overflow issue, but having a Spacer here as well will cause the available space to be divided evenly between the Flexible and the Spacer. Example:

Screenshot 2023-09-30 at 8 03 24 AM

I don't think you need the Spacer here so I would opt for removing it.

m-rosario (changes requested):

rupalg:
Screenshot 2023-10-02 at 6 16 22 PM
The commits above:
- revised the code based on the feedback requested above
- Revised Invitations section, with notification bubble. Show only the 3 most recent received invites.
- Revised Resources section (UI only)

Upcoming work:
- Revised Recommended section, recommending either Entrepreneurs or Mentors based on the authenticated user's group membership

lib/widgets/atoms/profile_header.dart
```diff
@@ -18,53 +19,68 @@ class ProfileHeader extends StatelessWidget {
...
+                    softWrap: true,
+                    style: theme.textTheme.titleLarge?.copyWith(
+                      color: theme.colorScheme.onSurface,
+                      fontWeight: FontWeight.w400,
+                    ),
+                  ),
                 ),
               ),
             ],
           ),
```
m-rosario, line 67: Flex only works for child widgets of a Column or Row. Putting this Flexible inside the Stack negates the flex, which again causes it to overflow:

Screenshot 2023-10-03 at 10 26 13 AM

I cannot think of any reason these widgets needs to be inside a stack, so maybe we can just rid of it.

lib/widgets/molecules/resources_section.dart
```diff
@@ -1,54 +1,82 @@
...
+      padding: const EdgeInsets.only(
+          top: Insets.paddingSmall, bottom: Insets.paddingSmall),
+      child: Column(
+        children: [
+          Image(image: resource_image),
+          Padding(
+            padding: const EdgeInsets.only(top: Insets.paddingSmall),
+            child: Text(subheading,
+                style: theme.textTheme.titleSmall?.copyWith(
+                  color: Colors.black,
```
m-rosario, line 30: Please don't use hardcoded colors. I know that this is what UX used, but we should know how to translate these to their equivalent color scheme value. **onSurface** is probably the best color role to use here.

lib/widgets/molecules/resources_section.dart
```diff
@@ -1,54 +1,82 @@
...
     List<Widget> resourceTiles = [
-      ImageTile(
-        image: const AssetImage(Assets.resourceWebinarStockImage),
-        title: l10n.homeResourcesWebinar,
-        subtitle: l10n.homeResourcesWebinarDescription,
-        isCircle: false,
+      Text(
+        l10n.moreResources,
+        style: theme.textTheme.titleLarge?.copyWith(
+          color: Colors.black,
```
m-rosario, line 49: No hardcoded colors.

lib/widgets/screens/dashboard/dashboard.dart
```diff
@@ -69,10 +67,6 @@ class _DashboardScreenState extends State<DashboardScreen>
                 MaybeReminderBanner(
                   authenticatedUser: _authenticatedUser,
                 ),
-                RecommendedSection(
-                  authenticatedUser: _authenticatedUser,
-                ),
```
m-rosario, line 74: Don't remove this until we have a replacement for it.

lib/widgets/atoms/section_tile.dart
```diff
@@ -107,8 +124,8 @@ class _SectionExpandToggle extends StatelessWidget {
         padding: const EdgeInsets.all(Insets.paddingExtraSmall),
         child: Text(
           text,
-          style: theme.textTheme.labelSmall?.copyWith(
-            color: theme.colorScheme.primary,
+          style: theme.textTheme.labelMedium?.copyWith(
+            color: Colors.black,
```
m-rosario, line 128: No hardcoded colors.

lib/widgets/atoms/section_tile.dart
```diff
@@ -2,9 +2,11 @@ import 'package:flutter/material.dart';
 import 'package:flutter_gen/gen_l10n/app_localizations.dart';
 import 'package:mm_flutter_app/constants/app_constants.dart';
 
+import 'notification_bubble.dart';
+
 class SectionTile extends StatefulWidget {
   final String title;
-  final Widget child;
+  final List<Widget> child;
```
m-rosario, line 9: SectionTile was meant to be a generic widget that could be customized into specific sections. However, making *child* a list of widgets and then using it's length to show a notification bubble is very specific to the invitation section. If you want to have this logic be available to all sections it should be more generic, if not then we prefer to have this logic implemented into its own widget.

If the InvitationSection is the only that will use the SectionTile then feel free to merge it to avoid clutter.

lib/widgets/atoms/section_tile.dart
```diff
@@ -42,19 +45,31 @@ class _SectionTileState extends State<SectionTile> {
...
-                widget.title,
-                style: theme.textTheme.titleSmall?.copyWith(
-                  color: theme.colorScheme.primary,
-                ),
+              Row(
+                children: [
+                  Text(
+                    widget.title,
+                    style: theme.textTheme.titleMedium?.copyWith(
+                      color: Colors.black,
```
m-rosario, line 61: Don't use hardcoded colors.

m-rosario (changes requested):

lib/widgets/screens/dashboard/dashboard.dart
rupalg: I will add a replacement here.

lib/widgets/atoms/profile_header.dart
rupalg: Hi Miguel, 
The reason I chose to introduce the "Stack" was because I wanted use mainAxisAignment.spaceBetween property to create space between the "Stack" of Positioned and Flexible widgets as well as the three dot menu on the right. Is there a better way to make sure the three dot menu is on the right in a row? I'll also continue to investigate and would appreciate insights.

lib/widgets/atoms/section_tile.dart
rupalg: Hi Miguel,
Yes, I see you had used SectionTile for More Resources cards before. Currently, the only thing that is using SectionTile is the invitation section. When you say "merge it", do you mean I should define a "SectionTile" class in the invitation_section.dart file? and revert the section_tile.dart to its original form?
Let me know!

lib/widgets/atoms/profile_header.dart
m-rosario: This can be accomplished easily with a Row I believe. Just place the Row children in the correct order from left to right and make sure that the child in the middle (the text) has a flex value so that it takes all of the available space, pushing the three dot menu to the right. Row can also use the spaceBetween property, but it won't be necessary if you use flex to get rid of space between widgets.

lib/widgets/atoms/section_tile.dart
m-rosario: I meant that we should either keep using SectionTile as a generic widget for all sections of the Home page, or make this logic you implemented here part of the InvitationSection, since none of the other sections would use it.

m-rosario: The idea was to have repeating elements of Home page sections encapsulated in the SectionTile, so that we don't have to define it multiple times. This is similar to how we used a template widget for the signup flow.

lib/widgets/atoms/profile_header.dart
rupalg: Thanks Miguel. Appreciate the insight. Will play with the flex value and report back

lib/widgets/atoms/section_tile.dart
rupalg: "make this logic you implemented here part of the InvitationSection" I will do the following, since I do not see elements other widgets needing SectionTile

rupalg:
Hi @m-rosario 
This is a draft and I have some more things to work on, and so no need to review yet. I have still committed because I have addressed your previous comments, if you wish you take a look.

m-rosario:
Thanks for all the hard work 🔥 

I forgot to mention that we also need the CTA in the reminder banner to navigate to the user's profile. That should be simple, just use the router to push the Profile page. 

We also need to implement the X button in the top right corner, which should hide the banner. It still hasn't been defined if the banner should remain hidden forever after dismissing it, so for now only close it when pressing X and don't worry about it appearing again when navigating back into Home.

rupalg:
Of course. Sounds great, Miguel! Will implement the needed functionalities, and send a commit here.

rupalg:
Made some progress, and will continue to post here when there is more progress. Feel free to merge it if you'd like, if wait it out.

rupalg:
@m-rosario got the UI to work, but need help with:
1) Connecting it to the backend queries
2) Getting the Divider to show up-- I cannot figure out yet why it is not showing up

m-rosario (approved):
Thanks, I will take care of the rest for the Home page.

PR #159, opened 2023-10-01
Inbox polling & improved Selectors

m-rosario:
This introduces additional improvements to the InboxModel:

- The redundant queries were consolidated to reduce data utilization.
- Polls the backend at regular intervals to retrieve new invitations received, new channels created, or determine if a sent invitation has been accepted or declined. This is necessary because subscriptions require an existing channel, and the events mentioned lack a channel to subscribe to. The scope of the poll will eventually be reduced to only include the most recent data to avoid wasteful data utilization.
- InboxModel Selectors were overriden to only rebuild widgets when certain model fields change, AND the model is not in a loading state. This is extremely important when polling because we do not want the screen to be rebuilding every time a poll occurs, only when the displayed elements change.

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -71,6 +71,12 @@ class _StartScreenState extends State<StartScreen> {
   }
 
   @override
+  void dispose() {
+    _inboxModel.cancelDataPolling();
+    super.dispose();
+  }
+
+  @override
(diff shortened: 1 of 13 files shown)


ashlilili (approved):
Thanks!

PR #160, opened 2023-10-02
Provide pronouns along with other options

andersan:
Fixes an issue which was present in #158 in the signup flow, where the pronoun options weren't loaded from the backend.


m-rosario (approved):
Thanks!

PR #161, opened 2023-10-03
Explore page improvements

m-rosario:
This PR adds several improvements to the Explore page, most notably the following:

1. The search query was refactored to remove it from the widget build method to avoid executing the search on every frame. This was causing the page to load slower. 
2. Polling is now done using the exponential backoff callback provided by the retryOnException function, up to 10 attempts before giving up.
3. Disable caching, for now. This was causing the search to return the same empty result over and over. We will definitely use caching in the future, but we need to have some logic in place to avoid issues like the one mentioned, which are extremely difficult to debug.
4. Fix expertises being displayed in the Industry dropdown. 
5. Update the Advanced Filters to the latest UX revision. This includes industry multi-select, single selection of user type, and business stages multi-select when Entrepreneur type is selected.
6. Added back and close buttons in the appbar for the filter pages.
7. Empty results now display a message to let the user know that no results were found.
8. Set a default user type filter when opening a fresh search. Entrepreneurs see a list of mentors, and vice-versa.
9. Fixed several UI overflows.
10. Allow closing drop-down menus by clicking outside of them.

Note: Pagination is still pending, since I do not see a way to skip results in the queries we are using. 15 results should be enough for the testing session though so this is not a huge problem.

[exploreV2.webm](https://github.com/micromentor-team/mm-flutter-app/assets/25093428/a8c50e0f-f3cc-41ab-afec-c2a8277426f6)


ashlilili (approved):

lib/providers/models/explore_card_filters_model.dart
```diff
@@ -80,14 +86,17 @@ class ExploreCardFiltersModel extends ChangeNotifier {
...
           _selectedLanguages.isEmpty ? null : _selectedLanguages.toList(),
-      offersHelp: _selectedUserTypes.contains("Mentor")
+      industriesTextIds:
+          _selectedIndustries.isEmpty ? null : _selectedIndustries.toList(),
+      companyStagesTextIds:
+          _selectedStages.isEmpty ? null : _selectedStages.toList(),
+      offersHelp: _selectedUserType == UserType.mentor
           ? Enum$UserSearchFieldPreference.isTrue
-          : Enum$UserSearchFieldPreference.any,
+          : Enum$UserSearchFieldPreference.isFalse,
```
andersan, line 95: We do have some users in our current app who are both mentors and mentees. With this logic, we'd never show those users, which is why I'd asked Michal to use the `any` option as a backup. 

Not a big deal for the moment, since we haven't designed any other features for users who have both entrepreneur and mentor roles yet. At some point we're likely to change this though

PR #162, opened 2023-10-03
Website signup -> optional, support "dual-role" users

andersan:
This PR contains 2 changes:
1. Allow entrepreneur/mentee users to continue through signup without entering a website
2. Show "dual-role" users (who are both mentors and mentees) in search results

For the optional input, I wasn't sure if we have a specific way to implement this.

2 isn't scoped for this app but it is a use case the MicroMentor will need to support, feel free to reject if it seems like it's outside the scope for the open source app. In that case I can revert that commit and merge this without that change.

### DIFF
--- a/lib/widgets/screens/sign_up/sign_up_entrepreneur_website.dart
+++ b/lib/widgets/screens/sign_up/sign_up_entrepreneur_website.dart
@@ -45,12 +45,10 @@ class _SignupEntrepreneurWebsiteScreenState
         leftOnPress: () {
           context.pop();
         },
-        rightOnPress: _website?.isNotEmpty ?? false
-            ? () {
-                _registrationModel.updateUserInput.companyWebsite = _website;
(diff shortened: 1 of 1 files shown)


m-rosario:
The website change looks good!

For dual-role search, I would prefer if this is done once the concept is fully defined and implemented in the app. Up to this point the requirement has been to assume you are either one type or the other, so the app logic is not designed to handle both. This might lead to indeterminate and confusing behavior such as the Mentor tag being displayed in one page, but Entrepreneur in the other. Meaning that you might search for a Mentor, but when you open the profile it says Entrepreneur, with mixed fields from both types. I would rather we resolve this carefully after the testing session.

andersan:
Thanks for the review, Miguel. I removed the dual-role change.

Agreed that it makes sense to wait to support the "dual role" user behavior once the requirements are clearer. There may be a lot of edge cases that come up as a result of that behavior.

PR #165, opened 2023-10-05
Add pagination support to the Explore page

jupblb:
This PR adds support for pagination to the Explore view. There are two issues:
1) Backend does not seem to respect the `skip` field of `Input$FindObjectsOptions`. @andersan 
2) For some reason the `onReady` function fires twice. The first time it does the `snapshot` value is set to the previous query result. The second time it fetches the actual result. I haven't yet figured out why it works like that. @m-rosario


m-rosario (approved):
Queries and network operations should not be in the build function, since this will be called several times by the framework when rendering the views. I submitted a commit to refactor the query out of build and do some general cleanup. Great job doing this, thanks for the help!

@andersan I tested the skip option and indeed it seems to be repeating results. The strange thing is that when skipping some results are new but others repeat, so I'm not sure what is going on here. Also if you try to skip beyond the number of results created by the createUserSearch query the backend will return the entire set of matches, which might be too big to handle. I added safeguards to indicate the "end of results" and avoid issues, but just thought I would mention this since it was unexpected behavior.

andersan:
Thanks for reporting. I've tracked the reported search bugs in the backend:
1. duplicate results/repeated users
2. if `skip` is greater than the `size` of the results, `limit` and `skip` are not applied

I had noticed one query that had one duplicate user out of 62 matches, but wasn't checking closely for this. If this is happening frequently, addressing this bug needs to be the top priority for search/matching on the backend.

Is there a use case where y'all consistently saw duplicate users? I just ran a search with 142 users (all mentor users in the US) and they were all unique if I return the full list. They also didn't have duplicate results in the first 2 pages of 10 users, but I didn't set up a more serious test yet.

m-rosario:
@andersan Check this example. I create a search with a max of 10 users, then get the results in groups of 3:

image
image
image
image

andersan:
@m-rosario thanks for sharing. I have been testing with larger lists (limit of 10), it's much easier to spot when limiting to 3 results. 

We probably need to add more sort filters here. I'll look into this early next week and follow up when this is fixed.

PR #166, opened 2023-10-06
Channel messages infinite scrolling with lazy loading

m-rosario:
Fetches the most recent messages of a chat whenever it is opened, and only fetch older messages as the user scrolls up. This allows opening the chat very quickly and reduce data utilization since it is not expected users need the entire history of chat messages every time they open a conversation.

This video demonstrates the feature with a conversation that has a very large number of messages:

[lazy_loading.webm](https://github.com/micromentor-team/mm-flutter-app/assets/25093428/6d30d4bc-1241-4379-b8b7-ffa56477f5a4)

### DIFF
--- a/lib/widgets/screens/inbox/channel_messages/channel_messages_screen.dart
+++ b/lib/widgets/screens/inbox/channel_messages/channel_messages_screen.dart
@@ -113,8 +113,6 @@ class _ChannelChatState extends State<ChannelChat>
   final ScrollController listScrollController = ScrollController();
   final Duration _animationDuration = const Duration(milliseconds: 250);
   late final ChatModel _chatModel;
-  int _messageCount = 0;
-  bool _unreadMessages = false; // Non-local Message exists outside of viewport
   ChannelMessage? _focusedMessage; // Intended reply Message
 
   @override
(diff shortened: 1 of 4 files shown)


lib/widgets/screens/inbox/channel_messages/channel_messages_screen.dart
```diff
@@ -123,65 +121,70 @@ class _ChannelChatState extends State<ChannelChat>
...
+    _chatModel.markMessagesAsRead();
+    if (_chatModel.channelMessages.last.createdBy ==
+        _chatModel.authenticatedUser.id) {
+      _scrollDown();
+    }
+  }
+
+  void _fetchOlderMessages() {
+    final extentAfter = listScrollController.position.extentAfter;
+    if (extentAfter < 300 &&
```
andersan, line 144: When does the scroll listener get triggered to load more messages? When there are less than 300 pixels between the top of the content (which is hidden) and the visible screen, we load 25 more messages?

Dug through the Flutter docs a bit (ScrollListener, ScrollPosition, the [extentAfter property](https://api.flutter.dev/flutter/widgets/ScrollMetrics/extentAfter.html)) to try to understand this, but I'm still not 100% sure I know what's happening here.

andersan (approved):
Looks good! It's working for me (a chat with more than 25 messages seems to seamlessly load earlier messages).

Had one question about when the additional messages are actually loaded, which I added ina. comment inline.

lib/widgets/screens/inbox/channel_messages/channel_messages_screen.dart
m-rosario: Yes that is pretty much what is happening. The scrollable content goes beyond the visible top of the screen, and this function triggers when there is less than 300 pixels to reach the top of the scrollable area. Once more messages load, the top of the scrollable area gets pushed up further so the user has to keep scrolling to trigger this function again.

The number 300 is sort of a magic number I arrived to when testing. Much lower than this and scrolling won't be smooth, much higher than this and the loading gets triggered too easily and sometimes more than once since the number of messages fetched doesn't push the scrollable area up enough.

PR #167, opened 2023-10-08
Edit profile page

ashlilili:
1. edit profile header when the profile belongs to the current user
2. about me section of edit profile page

image


m-rosario (approved):
Good stuff!! 

I made a commit to add the preferredLanguage field to the query. This is the field that is used in **Preferred Language**, while spokenLanguages goes in **Other Languages**. Sorry if this caused confusion, my bad!

PR #168, opened 2023-10-09
Cleanup 1st Pass: Delete unused GQL operations, organize widgets by feature

m-rosario:
- Deletes unused GraphQL operations and simplifies some of the used ones.
- Organize widgets by feature with 'screen' (accessed via router) widgets on the top level and components in a subfolder. Components that are shared by more than one feature are placed in a **shared** folder.


andersan:
This reorganization looks great. Can you ping me/us when you'd like to merge these into main and have no more local changes @m-rosario ? I didnt check anything in detail. Just looked it over briefly

m-rosario:
@andersan Thanks! The sooner I can merge this the better. I only had to add commits to resolve conflicts from a recent merge to main. Merging soon can help avoid more conflicts.

andersan:
Thanks Miguel.

Looking over [the doc Holger wrote up](https://github.com/micromentor-team/mm-flutter-app/blob/main/docs/Component-Structure.md#feature-based-architecture) about the feature based architecture vs atomic design, it references [this blog post](https://medium.com/flutter-community/scalable-folder-structure-for-flutter-applications-183746bdc320) as the source for the feature-based architecture idea. 

The proposed structure seems to be as follows:
```
--lib
  |--feature_1
  |  |--screens
  |  |--widgets
  |  |--models
  |  |--services
  |  |--view_models
  |--feature_2
  |  |--screens
  |  |--widgets
  |  |--models
  |  |--services
  |  |--view_models
  |....
```

IMO the structure in this PR is good. It doesn't seem necessary to match the above structure, and seems like that effort could be better spent elsewhere. Still, it does differ from the above structure. How did we end up with separating the assets into files with screens at the top level of each feature, and everything else in the `component` subdirectory?

Could you add a short note in that `Component-Structure.md` doc about this? Not a huge deal, but seeing how this is one of the few decisions documented in the repo, it makes sense to add something quick there.

m-rosario:
@andersan This is just the first pass and is definitively not the end goal. Think of it as an intermediate step between what we have now and what is documented there.

PR #170, opened 2023-10-11
Edit profile photo

ashlilili:
1. Add edit icon for profile photo if its the profile of the current user
2. updated the color of vacation banner to use primary95 - also created utility function for tonal palette

https://github.com/micromentor-team/mm-flutter-app/assets/100324646/dc25132c-1333-4ecc-9264-94b678b8a782


m-rosario (approved):
Thanks for creating the color utility and refactoring the dialog. Good stuff!

PR #171, opened 2023-10-12
Enable marking received invitations as read

m-rosario:
Received invitations are marked as read whenever the user opens the invitation detail page. At this point the inbox invite notification for this specific invitation is removed, regardless if the invite was not accepted or declined. I also added a plain notification bubble in the inbox received invitations list tiles whenever the invitation hasn't been opened.

Modified the mock server to support new field readByRecipientAt.

The following video shows the feature in action:

[received-invites-read-receipt.webm](https://github.com/micromentor-team/mm-flutter-app/assets/25093428/a6daa63e-3533-4d18-b2e7-42928442a5b7)

### DIFF
--- a/lib/schema/schema.graphql
+++ b/lib/schema/schema.graphql
@@ -299,6 +299,7 @@ type Query {
   findUserSearchResults(options: FindObjectsOptions, userSearchId: String!): [User!]!
   myUserSearches: [UserSearch!]!
   getMm2Integration: Mm2Integration!
+  findNlpConversation(conversationId: String!): NlpConversation!
   findMyActiveMultiStepAction: [SidMultiStepAction!]!
   getMultiStepActionProgress(
     """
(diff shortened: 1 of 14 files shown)


lib/schema/schema.graphql
```diff
@@ -359,6 +360,7 @@ type User {
   relationships: [UserRelationship!]
   inactivatedAt: DateTimeISO
   inactivatedBy: ID
+  termsAndConditionsAcceptedAt: DateTimeISO
```
andersan, line 363: Just an FYI - didn't realize this was already merged in the backend. I envisioned using this on [the signup/onboarding screen](https://github.com/micromentor-team/mm-flutter-app/blob/29350a623750e54bac3474f762c90b00960d859d/lib/widgets/features/sign_up/sign_up_method.dart) where the user accepts some conditions about MM.

Not sure if that's relevant given that we only run the `updateUser` graphql at the end of the signup, rather than on each screen.

andersan (approved):
Looks great, thank you Miguel!

PR #173, opened 2023-10-16
password_warnings

RaghvindYadav:
Refracting the signup password screen.


m-rosario:
FYI the UX for the password warnings was changed in the latest design revision, making most of these changes obsolete. The only requirement for passwords is a length of 8 characters or more. You could modify this PR to only include that requirement.

I suggest also rebasing to the newest version of the signup flow since the widget is much different now.

PR #175, opened 2023-10-16
Cleanup 2nd Pass: Organize English localizations by feature

m-rosario:
This is my honest attempt at cleaning up, renaming, and organizing over 300 localization strings. The naming convention is such that it goes from the most general to the most specific. They are also ordered alphabetically.

Deleted redundant and unused strings.

I believe it is now much easier to navigate this file and quickly identify all strings the belong to a specific feature or screen.

Note that I did not waste time trying to cleanup the Arabic and Spanish localizations since these were only used for a proof of concept and need to be redone completely in the future anyways.

HINT: The focus of this PR is the app_en.arb file, everything else is just refactoring to use the new names.


andersan (approved):
Looks great, thank you Miguel. Having these more descriptive names will be great for maintenance of this project going forward.

PR #176, opened 2023-10-16
Return pronouns from mock server

m-rosario:
Mock server was missing a resolver to return list of pronouns.

This should fix https://trello.com/c/Gx0SWLc3/132-bug-sign-up-pronoun-page-is-broken-in-the-sign-up and the related query from the edit profile page.


jonahtwalker (approved):
Thank you sir! 🙏🏼

PR #181, opened 2023-10-20
Push notification support

shadowbrush:
This also adds AppLifecycleListener to _StartScreenState to track when the app
is started and stopped and to report this to the backend, together with the
Firebase Messaging token.

OK, I admit I haven't done much testing on an actual device here. Feel free to tell me to go home with my PR and test that first.


andersan:
@shadowbrush thanks for these changes, they look great.

After running `firebase configure`, the app builds OK, but I am unable to use the signup flow. Logging in to an existing user works fine. I tried to fix that by catching an error in [this PR](https://github.com/micromentor-team/mm-flutter-app/pull/187), but it still fails for me in web. My iOS emulator is broken for some reason and i still haven't set up the Android emulator, so I'll do that now and re-test.

Is there any way to test the push notifications on an actual device with this setup? Or do we still need to do additional work on the backend or the distribution to do that?

m-rosario (approved):
Looks good 👍

andersan:
Emulators still both broken in my local, after updating xcode and messing with some of the config and minimum app version. If this is running fine in android/iOS it looks good to merge along with Miguel's fix in #188

PR #182, opened 2023-10-20
Signup UX Revision (2nd Pass)

m-rosario:
Further refines the signup flow according to the latest UX revision.

Key changes included here:

- Updated the Social Media SSO buttons in the signin and signup pages
- Added 'dots indicators' to the welcome page, which will be used in the future to change the slides showing a different picture and text. Right now these are clickable and change the index accordingly, but the slide doesn't change yet because we do not have the content for these other two slides.
- Update the Multi-select chips for the expertises. New version is much more compact and shows a checkbox for accessibility.
- Included the flutter_markdown package to allow including hyperlinks and rich formatting as part of the localization strings. This is important because different languages have different sentence structures, therefore formatting that is applied only to certain words needs to be included in the translations as well.
- Allowed pages in the signup flow to be scrolled vertically when there is not enough space in the screen (e.g. the keyboard is out). The navigation buttons remain on top always, and there is a subtle fade at the top of the screen for visual effect when scrolling.

Still pending: 

- field validation with corresponding error messages
- location search

The following video illustrates the entire flow with the changes above:
[signup-rev-2.webm](https://github.com/micromentor-team/mm-flutter-app/assets/25093428/a6aad5b9-10a6-439f-84e1-6674dcca18d8)


pubspec.yaml
```diff
@@ -69,6 +69,8 @@ dependencies:
   tuple: ^2.0.2
   collection: 1.17.1
   material_color_utilities: ^0.2.0
+  dots_indicator: ^2.1.1
+  flutter_markdown: ^0.6.18
```
andersan, line 73: https://github.com/search?q=repo%3Aflutter%2Fflutter+flutter_markdown&type=issues&p=2

Some issues to track on this package. I see it's a Flutter-built package, so there should be no issues using it, but when we get into less obvious functionality like a11y/screen readers, zooming/text resizing, etc. we need to double check that the markdown content works properly

andersan (approved):
Looks great, thank you Miguel.

PR #183, opened 2023-10-23
Update schema, search input changes

andersan:
Update to match changes to the matching engine on the explore and home screens


andersan:
@shadowbrush in case you're curious since you've been in the Flutter app recently, here's the updates to maintain compatibility with the matching engine's schema changes. Just changed the shape of the input, using an `Input$UserSearchFilterInput`. Running `flutter analyze` pointed out all the areas that needed to be changed after the GraphQL schema was updated.

m-rosario (approved):
Thanks!

PR #184, opened 2023-10-24
Edit Profile pages with backend integration

m-rosario:
This adds all remaining pages to edit profile fields. Profile is updated in the backend when the user navigates out of each individual edit page.

Note the following:

- Fields that are not working due to being missing in the backend: Business location, Mentoring preferences, additional expertises, photo upload
- As of the time this PR was created, there is a bug in the backend that clears all array fields in the mentor/mentee group memberships whenever the update mutation is called without including all optional parameters. This is a bug in the backend, so it is expected that this will be fixed without needing to update the frontend.
- I did some cleanup on the profile view to remove sections that were no longer needed (user prompts, expectations), and that were hardcoded (mentoring preferences, affiliations, business location). This is to facilitate testing and prevent confusion.
- City lookup is not yet implemented. As of the moment we are allowing the user to enter any value and it will be stored in the city field. Once lookup is implemented we will add the region and country fields to the mix.
- There is no field validation, and there is a lot more polishing to be done! Consider this a minimum product for the testing session.

The following video demonstrates how a minimal profile (created using the revised signup flow) can be edited and viewed:

[edit-profile.webm](https://github.com/micromentor-team/mm-flutter-app/assets/25093428/be7ed0a7-b155-46a3-9e02-7ad8c207404e)


andersan:
Laughed at the values for the video. Will check this out in detail in a bit, thanks Miguel!

lib/widgets/features/profile/components/edit_how_can_i_help.dart
```diff
@@ -92,9 +71,11 @@ class EditProfileHowCanIHelp extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    final topThreeExpertises = expertises.sublist(0, 3);
+    final topThreeExpertises = expertises.take(Limits.profileExpertiseMaxSize);
     final additionalExpertises =
-        (expertises.length > 3) ? expertises.sublist(3) : [];
+        (expertises.length > Limits.profileExpertiseMaxSize)
```
andersan, line 76: This is what i was thinking that we could do here, too, since it's not clear to me we that actually will retain these values. It also makes it easier for us to sync between our current app and the new one if we don't add a new field.

lib/widgets/features/profile/edit_profile_preferred_language.dart
```diff
@@ -36,31 +53,32 @@ class _EditPreferredLanguageScreenState
...
+            .toList(),
+        selectedOptions: _initialValue,
+      ),
+      editUserProfile: () => _userProvider.updateUserData(
+        input: Input$UserInput(
+          id: widget.userData.id,
+          preferredLanguageTextId: _preferredLanguagesController.getTags
+              ?.map(
+                (t) => _contentProvider.languageOptions!
+                    .firstWhere((o) => o.translatedValue! == t)
```
andersan, line 77: probably better to compare the `textId`s rather than the `translatedValue`s. in theory the translated value should work fine, but i don't know if it will work with every language (e.g. does comparing two arabic strings with `==` always work in Dart? no idea)

andersan (approved):
Fantastic job as always, Miguel. Thank you for getting this over the finish line before the testing session tomorrow.

lib/widgets/features/profile/components/edit_how_can_i_help.dart
m-rosario: It is an easy fix, but it is awkward to manage since both fields would be dealing with a single array and depending on the position of the elements to classify the as either 'top' or 'additional'. A lot of questions popped up with this approach, which is why I preferred to leave the 'additional' fields unimplemented for now.

andersan: Holger merged the PR i provided with the additional expertises fields. There was a lot of confusion about the requirements here though so leaving this un-implemented sounds good for now.

PR #185, opened 2023-10-24
Cleanup 3rd Pass: Restructure directories & use all relative imports

m-rosario:
Further organizes files into directories. Also changes all local imports to use relative imports, as per the Dart best practices recommendations.

The following illustrates the current directory structure:

image

Key changes in structure:
- Screens are now in their own folder under the feature folder (alongside **_components_**).
- The `providers` directory was moved to the `services/graphql/` directory, since these classes are exclusively used to interact with graphql queries and mutations.
- The `models` directory was moved out of the `providers` directory, and placed at the top level of the project. This is because these components interface with both widgets and services, meaning that they are very high level.
- The parts of the `constants` library have their own subfolder
- Firebase files are placed in a `firebase` directory under `services`.
- Cleaned up some files inside `utilities` by moving them to a more specific directory under the new structure.

All of this is being documented in the handoff doc which will be published once completed.


andersan (approved):
These refactors look great, thank you Miguel!

PR #187, opened 2023-10-26
Fix late variable error, relative import

andersan:
Minor changes - in the debugger there was an exception when `_tokenSubscriptionStream.cancel();` was called, which this fixes. Not sure if that has any effect on the app.

In my local, I am unable to view the app's onboarding flow after pressing "Get Started", which is what this PR was attempting to fix.


shadowbrush:
Great! Thanks for that, Chris! Let's merge this. I'm going to do some testing today, too.

shadowbrush:
Excited to see us do some Flutter work!

PR #188, opened 2023-10-26
Remove ShaderMask from signup flow to fix Web issues

m-rosario:
For some reason this widget makes it impossible to see the SignUp flow in the web version. Figured it was easier to remove it, since all it does is add a subtle fade effect at the top of the screen when scrolling vertically. No one will miss having this.


andersan (approved):
How'd you find this was the issue @m-rosario ? I didnt see any exception or anything related - was there some devtool you used or something with the debugger?

Great job!

PR #189, opened 2023-10-27
report user

saroj-kumar-padhi:


shadowbrush:
@saroj-kumar-padhi I'm making changes to your branch

lib/widgets/features/inbox/screens/inbox_invitation_detail.dart
```diff
@@ -208,7 +218,18 @@ class _InboxInvitationDetailScreenState
     AppLocalizations l10n,
     String senderId,
     String? senderName,
+    int selectedReason,
   ) {
+    final AppLocalizations l10n = AppLocalizations.of(context)!;
+
+    List<String> reasons = [
```
shadowbrush, line 225: We'll have to use different values that you don't have access to yet, Saroj. Let me prepare that.

shadowbrush:
I added `reportUser` to the backend. Let's use that instead of `createContentTag`. Under the hood `reportUser` creates a `ContentTag`, but the frontend does not need to be aware of that.

lib/widgets/features/inbox/screens/inbox_invitation_detail.dart
shadowbrush: I realized `declineChannelInvitation` does hot accept a `reason`, so I will add this it.

lib/widgets/features/inbox/screens/inbox_invitation_detail.dart
```diff
@@ -405,21 +463,25 @@ class _InboxInvitationDetailScreenState
 
 class DeclineReason extends StatefulWidget {
   final String? name;
+  int selectedReason;
```
shadowbrush: @saroj-kumar-padhi I'm no Flutter developer, but as I understand it you `DeclineReason` should be an immutable object and you should not add a state variable to it. Any user input should be saved into `_DeclineReasonState`.

shadowbrush:
@saroj-kumar-padhi For now, please make no changes to this PR as I'm still working on it. Please create a new branch and PR for the "block user" feature. Name the new branch something like `<your initials>/block-user`. That way we know who owns what branch.

shadowbrush:
I made some progress here, but it's still not done. But the app is now sending a reason to the backend when the user is declining an invitation. And those reasons are no longer hard-coded in the app, but are pulled from the backend, together with the translations.

PR #191, opened 2023-10-30
Minor mock server fixes: UserSearchInput & chat replies

m-rosario:
- Adds 'filter' to UserSearchInput in mock server, which was added to backend in https://github.com/micromentor-team/mm-flutter-app/pull/183. Without this the recommended section (Home) and the Explore page don't show any results.
- Remove reply comments from mock conversations, since replies are no longer supported by the UI (even though the code is still there for whenever this gets added again)


andersan (approved):
Looks good, thank you Miguel!

PR #192, opened 2023-10-31
Adapting server changing User.seeksHelp and offers to optional

shadowbrush:
This also applies the latest GraphQL schema.

Also:

* When searching, offersHelp and seeksHelp is never set to isFalse, rather to 'any'
* Users may not have fullName set. This uses firstName and lastName in this case.

### DIFF
--- a/lib/models/explore_card_filters_model.dart
+++ b/lib/models/explore_card_filters_model.dart
@@ -93,11 +93,11 @@ class ExploreCardFiltersModel extends ChangeNotifier {
             _selectedStages.isEmpty ? null : _selectedStages.toList(),
         offersHelp: _selectedUserType == UserType.mentor
             ? Enum$UserSearchFieldPreference.isTrue
-            : Enum$UserSearchFieldPreference.isFalse,
+            : Enum$UserSearchFieldPreference.any,
         searchText: selectedKeyword,
         seeksHelp: _selectedUserType == UserType.entrepreneur
             ? Enum$UserSearchFieldPreference.isTrue
-            : Enum$UserSearchFieldPreference.isFalse,
+            : Enum$UserSearchFieldPreference.any,
       ),
       maxResultCount: maxResultsCount,
     );
--- a/lib/widgets/features/explore/screens/explore.dart
+++ b/lib/widgets/features/explore/screens/explore.dart
@@ -148,8 +148,9 @@ class _ExploreCardScrollState extends State<ExploreCardScroll> {
           (user) => createProfileCardFromInfo(
             info: ProfileQuickViewInfo(
               userId: user.id,
-              userType:
-                  user.offersHelp == true ? UserType.mentor : UserType.entrepreneur,
+              userType: user.offersHelp == true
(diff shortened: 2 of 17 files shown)


lib/models/explore_card_filters_model.dart
```diff
@@ -93,11 +93,11 @@ class ExploreCardFiltersModel extends ChangeNotifier {
             _selectedStages.isEmpty ? null : _selectedStages.toList(),
         offersHelp: _selectedUserType == UserType.mentor
             ? Enum$UserSearchFieldPreference.isTrue
-            : Enum$UserSearchFieldPreference.isFalse,
+            : Enum$UserSearchFieldPreference.any,
         searchText: selectedKeyword,
         seeksHelp: _selectedUserType == UserType.entrepreneur
             ? Enum$UserSearchFieldPreference.isTrue
-            : Enum$UserSearchFieldPreference.isFalse,
+            : Enum$UserSearchFieldPreference.any,
```
m-rosario, line 100: Setting the fields to isFalse was done to avoid having users that are both mentors and entrepreneurs. I know that this is a valid use case in the backend, but the UX design for the frontend only supports one or the other.

Having a user with both memberships in the frontend means that whatever UI and functionality is displayed (mentor or entrepreneur) will be pretty much arbitrary, since it depends on which one of the two fields (seeksHelp/offersHelp) the developer checked first when implementing the logic.

TL;DR: The UX and business logic of the app needs to be revised if you plan to include dual membership users.

lib/widgets/features/explore/screens/explore.dart
```diff
@@ -148,10 +148,11 @@ class _ExploreCardScrollState extends State<ExploreCardScroll> {
...
             info: ProfileQuickViewInfo(
               userId: user.id,
-              userType:
-                  user.offersHelp ? UserType.mentor : UserType.entrepreneur,
+              userType: user.offersHelp == true
+                  ? UserType.mentor
+                  : UserType.entrepreneur,
               avatarUrl: user.avatarUrl,
-              fullName: user.fullName!,
+              fullName: user.fullName ?? '${user.firstName} ${user.lastName}',
```
m-rosario, line 155: Is there a way to have this logic in the backend so that instead of storing fullName as a separate field it would instead be calculated from the firstName and lastName?

lib/models/explore_card_filters_model.dart
shadowbrush: Thank you for pointing this out, Miguel. MicroMentor has dual role users, so we will have to support it. But this may not need to happen in the open source app - not my decision. I will close this PR, we will apply this change to the MicroMentor app and then discuss how it will be pushed upstream. 

Same with the handling of `fullName`, `firstName` and `lastName`. Of course, you are right to point out this could happen in the backend. The MicroMentor backend isn't using `fullName` at all, so we'll change it there in the app, and then have to discuss again how this affects the open source app.

shadowbrush:
Closing this PR as it brings MicroMentor specific changes.

PR #193, opened 2023-10-31
Initialize and maintain state during signup flow

m-rosario:
This PR makes it so that users can navigate back and forward in the signup flow pages without losing the values that were previously entered.

This is done primarily in two ways:

1. PageRoutes in the router are set to maintain state, which means that once created the screens will remain active and maintain the state of their variables even if there is another page in the top of the navigation stack. This helps when navigating to previous pages. * Note that performance is not a concern here because there can only be one copy of each signup screen in the navigation stack, and the stack is cleared once signup completes.
2. The Stateful Widget screens are initialized using values from the UserRegistrationModel, if any. This helps when navigating forward to a page you had filled out already, given that said page had to be popped and disposed of in order to get to a previous page, meaning that the maintain state method discussed above won't help in this case.

The following video demonstrates the feature:

[signup-save-state.webm](https://github.com/micromentor-team/mm-flutter-app/assets/25093428/2e06f6af-6120-4ee9-9ca0-46d5e56f4e3d)

### DIFF
--- a/lib/utilities/router.dart
+++ b/lib/utilities/router.dart
@@ -142,7 +142,7 @@ class AppRouter {
           pageBuilder: (BuildContext context, GoRouterState state) {
             return MaterialPage(
               key: state.pageKey,
-              maintainState: false,
+              maintainState: true,
               child: const SignupCredentialsScreen(),
             );
           },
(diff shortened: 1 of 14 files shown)


andersan (approved):
Thanks for the awesome work as always, Miguel. This was a UX issue that was easy to overlook but which can cause big issues for users, especially with the free text fields.

PR #194, opened 2023-11-01
Misc fixes: Android back navigation, README update, UI tweaks

m-rosario:
- Per UX request, make entire radio button card (used during signup flow user type and business stage selection) clickable. Adjust paddings.
- Fixes issues with handling the Android device's back button on API 34+, which would always cause the app to be closed instead of navigating to the previous page. The fix involves removing an entry from the Android manifest, which is suggested in [this issue](https://github.com/flutter/flutter/issues/117061).
- Remove sections on Widgetbook and testing from the README, since the app currently doesn't support these. This was done to avoid confusing new engineers and partners reading the document.


shadowbrush (approved):
Looks good, and thanks, Miguel!

